PR #30, opened 2023-05-03
Improve model description support

diegoferigo:
- Jaxsim can now load both SDF and URDF through `rod`. URDF support depends on the system availability of either Gazebo Classic or Gazebo Sim used for the conversion.
- The kinematic tree of the parsed SDF and URDF description is automatically converted to use poses compatible with the URDF format (i.e. having all links directly attached to their parent joint). This is the only supported format in URDF. SDF models not compatible are not currently supported. This is necessary since jaxsim assumes URDF compatibility when it builds a physics model.

cc @fl-ferr


traversaro (approved):

diegoferigo:
@fl-ferr I added also you as reviewer, can you please check that this works in your case even without the need of manually switching the frame convention to Urdf? Now it is done internally.

fl-ferr (approved):

fl-ferr (approved):

PR #34, opened 2023-06-09
Rewrite soft-contacts algorithm

diegoferigo:
- For maintainability purpose, the algorithm has been rewritten to operate on an individual point, and then applied model-wise through `jax.vmap`
- The computation of the position and velocities of the collidable points now does not longer use removed `jax.experimental` resources

The theory of the contact model can be found in [diegoferigo/phd-thesis](https://github.com/diegoferigo/phd-thesis):
- Formulation in Section 7.3
- Algorithm in Section 8.1.8


diegoferigo:
cc @fl-ferr

PR #35, opened 2023-06-13
Implement ABA with `jax.lax.scan`

fl-ferr:
* Remove `jax<0.3.16` pin

* Remove `jax.experimental.loops`

I wasn't able to remove the dependency on  `jax_dataclasses < 1.4.0` as it resulted in a circular import in the [`high_level`](https://github.com/ami-iit/jaxsim/tree/main/src/jaxsim/high_level) module.

C.C. @traversaro @CarlottaSartore

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -57,10 +57,10 @@ install_requires =
     coloredlogs
     distrax
     flax
-    jax >=0.3.14, <0.3.16
-    jaxlib == 0.3.15
+    jax 
+    jaxlib 
     jaxlie
-    jax_dataclasses >= 1.2.2, < 1.4.0
+    jax_dataclasses < 1.4.0
     pptree
     rod
     scipy
--- a/src/jaxsim/high_level/model.py
+++ b/src/jaxsim/high_level/model.py
@@ -2,7 +2,7 @@ import dataclasses
 import pathlib
 from typing import Any, Dict, List, Optional, Tuple, Union
 
-import jax.experimental.ode
+import jax
 import jax.numpy as jnp
 import jax_dataclasses
 import numpy as np
@@ -721,7 +721,7 @@ class Model(JaxsimDataclass):
         return f_B, tau
 
     def forward_dynamics(
-        self, tau: jtp.Vector = None, prefer_aba: float = True
+        self, tau: jtp.Vector = None, prefer_aba: bool = True
     ) -> Tuple[jtp.Vector, jtp.Vector]:
         return (
             self.forward_dynamics_aba(tau=tau)
--- a/src/jaxsim/physics/algos/aba.py
+++ b/src/jaxsim/physics/algos/aba.py
@@ -1,7 +1,6 @@
 from typing import Tuple
 
 import jax
-import jax.experimental.loops
 import jax.numpy as jnp
 
 import jaxsim.typing as jtp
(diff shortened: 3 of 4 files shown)


traversaro:
cc @diegoferigo

traversaro:
> I wasn't able to remove the dependency on `jax_dataclass < 1.4.0` as it resulted in a circular import in the [`high_level`](https://github.com/ami-iit/jaxsim/tree/main/src/jaxsim/high_level) module.

Thanks @fl-ferr ! Can you report the exact error you obtained with a recent jax_dataclass ?

fl-ferr:
The following error appears when trying to import jaxsim:

```python
No GPU/TPU found, falling back to CPU. (Set TF_CPP_MIN_LOG_LEVEL=0 and rerun for more info.)
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/home/flferretti/mambaforge/envs/test103/lib/python3.10/site-packages/jaxsim/__init__.py", line 63, in <module>
    from . import high_level, logging, math, sixd
  File "/home/flferretti/mambaforge/envs/test103/lib/python3.10/site-packages/jaxsim/high_level/__init__.py", line 1, in <module>
    from . import common, joint, link, model
  File "/home/flferretti/mambaforge/envs/test103/lib/python3.10/site-packages/jaxsim/high_level/joint.py", line 12, in <module>
    class Joint(JaxsimDataclass):
  File "/home/flferretti/mambaforge/envs/test103/lib/python3.10/site-packages/jax_dataclasses/_dataclasses.py", line 47, in pytree_dataclass
    return wrap(cls)
  File "/home/flferretti/mambaforge/envs/test103/lib/python3.10/site-packages/jax_dataclasses/_dataclasses.py", line 38, in wrap
    return _register_pytree_dataclass(dataclasses.dataclass(cls, **kwargs))
  File "/home/flferretti/mambaforge/envs/test103/lib/python3.10/site-packages/jax_dataclasses/_dataclasses.py", line 76, in _register_pytree_dataclass
    type_from_name = get_type_hints_partial(cls, include_extras=True)  # type: ignore
  File "/home/flferretti/mambaforge/envs/test103/lib/python3.10/site-packages/jax_dataclasses/_get_type_hints.py", line 80, in get_type_hints_partial
    value = eval(value, base_globals)
  File "<string>", line 1, in <module>
AttributeError: partially initialized module 'jaxsim' has no attribute 'high_level' (most likely due to a circular import)
```
The thing I noticed is that:

https://github.com/ami-iit/jaxsim/blob/2e50bf1f8e9b8b3b985c77fd9a49890062715391/src/jaxsim/high_level/__init__.py#L1

https://github.com/ami-iit/jaxsim/blob/2e50bf1f8e9b8b3b985c77fd9a49890062715391/src/jaxsim/high_level/joint.py#L5

https://github.com/ami-iit/jaxsim/blob/2e50bf1f8e9b8b3b985c77fd9a49890062715391/src/jaxsim/high_level/link.py#L5

There is an actual circular import, but I don't understand why the error is not raised with the old `jax_dataclass` version. I'm quite sure I'm missing something

traversaro:
Thanks, now the error is more clear, anyhow it is indeed quite indipendent from the issue that this PR fixes!

fl-ferr:
I think I found it! It should be related to the deprecation of the function `jdc.static_field()` (see https://github.com/brentyi/jax_dataclasses/commit/cba8ac9b844d906ac8d660f00e1ae9dbb695f322). I will create another PR

fl-ferr:
I did some tests and I noticed that I obtain two different results using the commits prior to 2e50bf1f8e9b8b3b985c77fd9a49890062715391. The two outputs below are related to the simple case of [free falling spheres](https://github.com/ami-iit/element_rl-for-codesign/blob/64db1c9253dd059de35cfb6a454e25a9dda8858f/src/examples/JAXsim_toyproblems/bouncing_ball_parallel.py) with the same ground and environment parameters:

Prior to rewrite of `soft_contacts`:
output_old

Current `jaxsim` version:
output_new


Notice that in both the cases I'm using the new ABA implementation.

traversaro:
@fl-ferr if you want to debug the two ABA implementations, probably it is easier to just check a single run of the ABA against a "good" reference implementation of the forward dynamics. In this way, you do have the complexity of the integrator running, and you can concentrate on debugging the ABA itself. 

If you have a reliable way of computing the Inverse Dynamics and the Mass Matrix, you can easily compute the Forward Dynamics computed by ABA as:

$$
\dot{\nu} = M(q)^{-1} \left(\tau + \sum_i J_i^T \mathrm{f}_i - C(q,\nu) \nu - G(q)\right) = FD(q, \nu, \tau, \mathrm{f}_i)
$$

See for example how ADAM does its tests: https://github.com/ami-iit/ADAM/blob/main/tests/test_NumPy_computations.py (even if in that case the reference implementation comes from another library, iDynTree in this case).

diegoferigo:
Thank you a lot @fl-ferr, I extended your modifications in #38, that superseeds this PR. I added you as co-author of the commits that were taken from here. Closing.

Issue #37, opened 2023-06-16
Update `step_over_horizon` to preserve `velocity_representation` attribute


fl-ferr:
When invoking the[ `step_over_horizon`](https://github.com/ami-iit/jaxsim/blob/2e50bf1f8e9b8b3b985c77fd9a49890062715391/src/jaxsim/simulation/simulator.py#L384-L470) method of the [`JaxSim`](https://github.com/ami-iit/jaxsim/blob/2e50bf1f8e9b8b3b985c77fd9a49890062715391/src/jaxsim/simulation/simulator.py#L52-L470) class the attribute `velocity_representation` is modified. 

Prior to the method call, `velocity_representation` is an object of type [`<Velrepr.Body: 1>`](https://github.com/ami-iit/jaxsim/blob/2e50bf1f8e9b8b3b985c77fd9a49890062715391/src/jaxsim/high_level/common.py#L4-L11). However, after executing [`step_over_horizon`](https://github.com/ami-iit/jaxsim/blob/2e50bf1f8e9b8b3b985c77fd9a49890062715391/src/jaxsim/simulation/simulator.py#L384-L470) it is changed to `DeviceArray(1, dtype=int64, weak_type=True)`.

It follows that when a new model is added using the method [`insert_model_from _description`](https://github.com/ami-iit/jaxsim/blob/2e50bf1f8e9b8b3b985c77fd9a49890062715391/src/jaxsim/simulation/simulator.py#L222-L262) it inherits the `velocity_representation` of the simulator, making impossible to call the [`Model.base_velocity()`](https://github.com/ami-iit/jaxsim/blob/2e50bf1f8e9b8b3b985c77fd9a49890062715391/src/jaxsim/high_level/model.py#L534-L542) method (returns a Value Error) and thus impossible to simulate.

C.C. @diegoferigo

fl-ferr:
Honestly I am not sure how correct my solution is, yet changing the `velocity_representation` field in the [`jax_dataclasses`](https://github.com/brentyi/jax_dataclasses) field as static in [line 65](https://github.com/ami-iit/jaxsim/blob/2e50bf1f8e9b8b3b985c77fd9a49890062715391/src/jaxsim/simulation/simulator.py#L65C65-L65C65) solved the problem:

```python
@jax_dataclasses.pytree_dataclass
class JaxSim(JaxsimDataclass):
    """The JaxSim simulator."""

    #[...]

[...]

PR #42, opened 2023-06-30
Allow keeping the base pose when a model is reduced

diegoferigo:


fl-ferr:
I've tried with my humanoid class for PPO and I've encountered a problem related to the new ABA version in the following line:

https://github.com/ami-iit/jaxsim/blob/0eab2861468cf852dbabbd83479d4d735d9d01d0/src/jaxsim/physics/algos/aba.py#L98

In fact respect to the one presented in #35 I've seen that you've removed a part of the code:

```diff 
1c1
< vJ = S[i] * qd[i - 1] if qd.size != 0 else S[i] * 0 
#35
---
> vJ = S[i] * qd[ii]
#42
```

This should allow to treat the case in which the model has no joints, e.g. in the case of this PR, when every joint is blocked

fl-ferr (changes requested):
I've tried with my humanoid class for PPO and I've encountered a problem related to the new ABA version in the following line:

https://github.com/ami-iit/jaxsim/blob/0eab2861468cf852dbabbd83479d4d735d9d01d0/src/jaxsim/physics/algos/aba.py#L98

In fact respect to the one presented in #35 I've seen that you've removed a part of the code:

```diff 
1c1
< vJ = S[i] * qd[i - 1] if qd.size != 0 else S[i] * 0 
#35
---
> vJ = S[i] * qd[ii]
#42
```

This should allow to treat the case in which the model has no joints, e.g. in the case of this PR, when every joint is blocked

diegoferigo:
Thanks, I didn't test this case. Should be fixed now, let me know if you still have problems.

fl-ferr:
Thanks for your changes Diego! I'm now encoutering a small issue that I described in https://github.com/ami-iit/jaxsim/issues/37. In particular, after having reduced the model, when I try to simulate using `step_over_horizon` the attribute `VelRepr` is changed to be a DeviceArray. As I said in https://github.com/ami-iit/jaxsim/issues/37#issuecomment-1604094101, making `velocity_representation` in the `JaxSim` class solved the issue for me

diegoferigo:
> Thanks for your changes Diego! I'm now encoutering a small issue that I described in #37. In particular, after having reduced the model, when I try to simulate using `step_over_horizon` the attribute `VelRepr` is changed to be a DeviceArray. As I said in [#37 (comment)](https://github.com/ami-iit/jaxsim/issues/37#issuecomment-1604094101), making `velocity_representation` static in the `JaxSim` class solved the issue for me

Mmh, I guess that by JIT-compiling a `JaxSim` class, jax wants to treat the enum of the velocity representation as a parameter that can be changed without triggering recompilation.

This is not something we want to achieve in the short term since it would involve using `jax.lax.select` everywhere in `high_level.model.Model`. Since [it's already static](https://github.com/ami-iit/jaxsim/blob/0eab2861468cf852dbabbd83479d4d735d9d01d0/src/jaxsim/high_level/model.py#L110) in `Model`, I guess it makes sense to keep it as static also in the simulator.

fl-ferr:
Thank you for your feedback! Looks good to me

fl-ferr (approved):

fl-ferr (approved):

PR #44, opened 2023-07-27
Fix trace leak by introducing a new pattern for jit-compiling class methods

diegoferigo:


fl-ferr (approved):

diegoferigo:
Given that the OOP decorators alter considerably downstream code and we're not yet sure they are 100% compatible with our long-term goals, I'll proceed with caution by merging this and next PRs into the `new_api` feature branch.

If everything keeps looking good, the new APIs will become the default ones.

PR #46, opened 2023-08-04
Support Python ≥ 3.10

### LINKED ISSUE #39: Support only Python ≥ 3.10
For the first release on jaxsim, I was considering to remove the support of python 3.8 and 3.9. It will bring us better dataclasses support (especially much improved inheritance) and the match-case statement.

I leave this issue here in case someone has valid points to keep older Python versions.

diegoferigo:
Closes #39


traversaro (approved):

traversaro:
fyi @fl-ferr

Issue #43, opened 2023-07-10
The simulator leaks JAX traces


diegoferigo:
After the major jax update merged in #38, compiling in jit some method raises [`jax.errors.UnexpectedTracerError`](https://jax.readthedocs.io/en/latest/errors.html#jax.errors.UnexpectedTracerError). Probably recent version of jax became more picky about it.

It's worth noting that the computation is correct if the methods are not jit-compiled. The leak error occurs only when a jax transformation is applied.

A MRE is the following:

```python
import pathlib

import robot_descriptions.double_pendulum_description

import jax
import jaxsim
from jaxsim.simulation.ode_integration import IntegratorType
from jaxsim.simulation.simulator import VelRepr

sim = jaxsim.JaxSim.build(
    step_size=0.001,
    steps_per_run=1,
    velocity_representation=VelRepr.Inertial,
    integrator_type=IntegratorType.EulerSemiImplicit,
).mutable(mutable=True, validate=False)

urdf_model = pathlib.Path(robot_descriptions.double_pendulum_description.URDF_PATH)

model = sim.insert_model_from_description(model_description=urdf_model)

# The problem occurs only if we jit-compile the method
sim.step = jax.jit(sim.step)

step_data = sim.step()

print(step_data)
```

<details>
<summary>Error</summary>

```

jaxsim on  main via 🐍 v3.10.12 🅒 /jaxsim  
✦ ❯ JAX_CHECK_TRACER_LEAKS=1 python /tmp/mre.py
jaxsim[109057] INFO Enabling JAX to use 64bit precision
Initialized persistent compilation cache at /tmp/jaxsim_compilation_cache
jaxsim[109057] DEBUG Using JAX compilation cache located at '/tmp/jaxsim_compilation_cache'
/jaxsim/lib/python3.10/site-packages/flax/struct.py:132: FutureWarning: jax.tree_util.register_keypaths is deprecated, and will be removed in a future release. Please use `register_pytree_with_keys()` instead.
  jax.tree_util.register_keypaths(data_clz, keypaths)
/jaxsim/lib/python3.10/site-packages/flax/struct.py:132: FutureWarning: jax.tree_util.register_keypaths is deprecated, and will be removed in a future release. Please use `register_pytree_with_keys()` instead.
  jax.tree_util.register_keypaths(data_clz, keypaths)
jaxsim[109057] DEBUG Found model '2dof_planar' in SDF resource
[...]

PR #48, opened 2023-08-04
Apply new OOP pattern compatible with jax transformations

diegoferigo:
This PR applies the resources of #44 on the jaxsim classes, solving #43 that is the original problem that triggered the development of this new approach.

Beyond fixing trace leaks, this PR:

- Enforces pytrees to be frozen, but enables developing r/w methods that succeed only if the pytree structure is not altered (unless manually allowed).
- Allows developers to catch early mistakes that would lead to cryptic jit recompilations.
- Allows users to develop applications with traditional OOP, with a much simplified approach to apply `jax.jit` and `jax.vmap`.
- Updates all returned types to be `jax.numpy` arrays, even when they are scalar quantities (in this case they are 0-dimensional arrays).
- Removes the confusing usage of `Mutability` from the methods, now the mutable/frozen context is enforced in the new decorators.

The key resources of this pattern are new method decorators and a `Vmappable` inheritance (introduced in #44). There are, however, some caveats to consider:

- Methods with arguments or returned object of types not supported by jax transformations (e.g. `str`) cannot be jit compiled nor vectorized.
- Default arguments of vmappable methods cannot use the scalar value. They have to be initialized as `None` and the default value has to be configured inside the method.


diegoferigo:
I got back working on this PR. The new OOP decorator works already pretty good for r/o methods, however the r/w logic miserably fails.

Our OOP strategy for the high-level resources (`Model`, `Link`, `Joints`) consists of the following design:

- The entire state is stored inside [`Model.data`](https://github.com/ami-iit/jaxsim/blob/eeec516b98da3a05586045c60c313986559fa2a1/src/jaxsim/high_level/model.py#L110).
- Links and joints objects are created and returned on the fly by [jit-compiled methods](https://github.com/ami-iit/jaxsim/blob/eeec516b98da3a05586045c60c313986559fa2a1/src/jaxsim/high_level/model.py#L398-L437).
- Links and joint objects contain a reference to their [parent model](https://github.com/ami-iit/jaxsim/blob/eeec516b98da3a05586045c60c313986559fa2a1/src/jaxsim/high_level/link.py#L28-L34) so that they can read -and alter- the state (i.e. `data`).

The main problem of this first version of the design is that a jit-compiled -decorated- r/w method of links and joints (like [`Link.add_external_force`](https://github.com/ami-iit/jaxsim/blob/eeec516b98da3a05586045c60c313986559fa2a1/src/jaxsim/high_level/link.py#L253-L291)) applies a [simple logic to copy the pytree's dynamic leafs](https://github.com/ami-iit/jaxsim/blob/eeec516b98da3a05586045c60c313986559fa2a1/src/jaxsim/utils/oop.py#L195-L217) from the object returned by the associated functional execution (`obj`) to the original OO `instance`.

When this logic runs, the `obj._parent_model` attribute has a different `id` than `instance._parent_model`, and the reference get lost. It seems to me that  discusses similar problems and use cases.

I've been experimenting with few solutions similar to those proposed in https://github.com/google/jax/issues/7919 and https://github.com/google/jax/discussions/17341, but the logic seems complicated to maintain and quite error prone.

**In a context of data sharing** (that could be state like in this case, but also more generic r/w parameters), probably the most simple solution is **not to allow any r/w methods on child classes** like `Link` and `Joint`. We have just few of them, that can be moved to `Model` for now. I'll follow with interest possible upstream development of ideas from other jaxsim users.

Note that we have a similar relationship also between `JaxSim` and `Model`, but in this case there's no parameter sharing. Therefore, having r/w methods in `Model` -that operate directly on its `data` attribute- should be ok.

diegoferigo:
> Happy to help adapting my class as shown in [google/jax#7919 (comment)](https://github.com/google/jax/issues/7919#issuecomment-1711475155) I was able to convert an OOP code base with this and have it work in a metagradient descent scenario. Key is providing a flatten/unflatten function for every class but the class ReffableTreeNode does a lot of the complex work to make this super easy to implement...

Thank you @samskiter for chiming in! I'm definitely interested in your approach using `ReffableTreeNode`. For the moment, I worked around the limitation described in https://github.com/ami-iit/jaxsim/pull/48#issuecomment-1746635121 by removing r/w methods of classes operating on data not owned by them. I found that r/o methods are, of course, working fine, but problems occur in case of jitted r/w methods.

Luckily, we only have two of such methods, and I moved them for now to the parent class (https://github.com/ami-iit/jaxsim/pull/48/commits/5a975d4f6afa28d94e89ccd6576022be7586c06f).

I keep in mind to try playing around with your solution, it could be an excellent companion to our [`JaxsimDataclass`](https://github.com/ami-iit/jaxsim/blob/new_api/src/jaxsim/utils/jaxsim_dataclass.py) and [`Vmappable`](https://github.com/ami-iit/jaxsim/blob/new_api/src/jaxsim/utils/vmappable.py). Is the project where you use it open source? If yes, I'd like to have a look to your use cases. If not (yet), I'll be following your projects.

flferretti (approved):
LGTM

diegoferigo:
> Afraid not open source, but I'm happy for you to use ReffableTreeNode - the toy usage in my other comment still stands.

Yeah I didn't find anything in your repos and I assumed still being closed source. I opened #52 for collecting resources and attempts we might perform in the future. For now, thanks a lot for the details!

PR #54, opened 2023-10-10
First steps towards automatic differentiation of RBDAs

diegoferigo:
This PR works around some problems I encountered when trying to make automatic differentiation (AD) run on our rigid-body dynamics algorithms (RBDAs).

This is not the first time I've tried this in the past, but every time there were `NaN`s generated somewhere in the code that were quite difficult to spot. I never had enough time to dedicate to dissect small sections and proceed with a pragmatic debugging campaign.

Luckily, JAX development progressed considerably, and now thanks to the following features:

- better errors in case of problems,
- `JAX_DEBUG_NANS=1` that points directly to the origin of any NaN ,
- `JAX_DISABLE_JIT=1` that disabled JIT entirely also inside e.g. `jax.lax.scan`, allowing to spot NaNs occurring there,

I finally managed to make the tests against finite difference (with `jax.test_util.check_grad`) work, both for the first-order and second-order derivatives! :tada: I guess we are one step closer to a fully differentiable simulator, and definitely towards #4.

For the moment, this PR checks that JAX is able to differentiate through the following algorithms:

- ABA
- RNEA
- CRBA
- Forward kinematics
- Jacobians
- Soft contacts

I still have to check properly our integrators in a future PR. I suspect that the integration of quaternion with Baumgarte stabilization might be suboptimal in these cases, but it can be simply solved by implementing the integration on SO(3). I'm considering giving it a try.

### DIFF
--- a/src/jaxsim/math/quaternion.py
+++ b/src/jaxsim/math/quaternion.py
@@ -33,7 +33,7 @@ class Quaternion:
         omega_in_body_fixed: bool = False,
         K: float = 0.1,
     ) -> jtp.Vector:
-        w = omega.squeeze()
+        ω = omega.squeeze()
         quaternion = quaternion.squeeze()
 
         def Q_body(q: jtp.Vector) -> jtp.Matrix:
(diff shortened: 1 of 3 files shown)


DanielePucci:
CC @Giulero

flferretti (commented):
That's great! Yet it generates NaNs for the free falling example using the hopper model after $\sim1.2s$:

ad_hopper

flferretti (approved):

PR #58, opened 2023-11-09
Fix CI/CD

diegoferigo:
The pins are only necessary in the current version of `main`. The upcoming [`new_api`](https://github.com/ami-iit/jaxsim/tree/new_api) branch should already be compatible with the latest jax versions.


flferretti (approved):
LGTM

PR #65, opened 2024-01-15
Add JAXsim Notebook for Parallel Simulation

flferretti:
This PR will add a notebook with [pixi](https://github.com/prefix-dev/pixi) support as a guide to write parallel simulation with JAXsim.

In particular, this example will concern the simulation of multiple free falling spheres, with different initial pose.

The idea is to avoid the free-fall motion visualization for this example, but rather to plot the height time serie for each sphere, as I don't know if MuJoCo supports multiple models rendering.


diegoferigo (approved):
LGTM with few comments:

- Maybe it's worth mentioning that parallel simulations are independent from each other. Adding a pose displacement is just useful to visualize a parallel simulation in a single visualizer instance without having overlapping models.
- Can you check if the usage of `mutable()` is necessary? With the new OOP decorators, the mutable methods should already take care of enabling the mutability.
- Probably is also worth specifying that `step_over_horizon` is useful only in the open-loop case. It can be used also in a closed-loop case, but it is much more complicated. At least in the beginning, using the regular `step` might be preferred.

flferretti:
Thanks @diegoferigo, solved in daa5a9e and 2598d24. Merging

PR #67, opened 2024-01-24
Prepare for first release

diegoferigo:
- Support Python ≥ 3.11. The upcoming Ubuntu 24.04 LTS has this version, it makes sense to start using the new features from the beginning.
- Updated the minimum supported version of dependencies. For `jax`, particularly, let's try to keep a good upstream alignment since the availability of new features might be useful on the long term.
- Enabled windows in CI similarly to macOS, where the package is only imported without running the test suite.

Some comment:
- Let's follow the usual semantic versioning https://semver.org/.
- Let's follow [PEP440](https://peps.python.org/pep-0440/) (https://packaging.python.org/en/latest/specifications/version-specifiers/#version-specifiers).

Notes:
- On the long run, we can consider to ditch the `setup.cfg` and include all the information into `pyproject.toml`.


flferretti (approved):
LGTM

diegoferigo:
PR ready for review. Required to push https://github.com/conda-forge/staged-recipes/pull/24630 through.

We can add the conda installation instructions in the new main (will be included in the next release, not a big deal).

flferretti:
@diegoferigo we should bump also this https://github.com/ami-iit/jaxsim/blob/main/src/jaxsim/version.txt or find a different solution

diegoferigo:
I missed that file, I had no idea it was there. We can generate it with [`setuptools_scm`](https://github.com/pypa/setuptools_scm), let me check.

flferretti:
What about putting in [`jaxsim/src/jaxsim/__init__.py`](https://github.com/ami-iit/jaxsim/blob/main/src/jaxsim/__init__.py):
```python
__version__ = "0.1.0"
```

diegoferigo:
Would that be enough for readthedocs? I don't like hardcoded data since there is almost 100% chance to forget to update it. I just checked, `setuptools_scm` writes the following `jaxsim/_version.py` file:

```python
jaxsim on  first_release via 🐍 v3.11.7 🅒 /jaxsim  
❯ cat src/jaxsim/version.py
# file generated by setuptools_scm
# don't change, don't track in version control
TYPE_CHECKING = False
if TYPE_CHECKING:
    from typing import Tuple, Union
    VERSION_TUPLE = Tuple[Union[int, str], ...]
else:
    VERSION_TUPLE = object

version: str
__version__: str
__version_tuple__: VERSION_TUPLE
version_tuple: VERSION_TUPLE

__version__ = version = '0.1.dev398+dirty'
__version_tuple__ = version_tuple = (0, 1, 'dev398', 'dirty')
```

If readthedocs is not capable of reading this file, we can just add the following in the `jaxsim/__init__.py`:

```
# [...]
from ._version import __version__
```

Anyway, is this necessary? Have you tried https://stackoverflow.com/a/61175918?

flferretti:
Maybe we could opt for something like this in `conf.py`:
```python
import requests

# Make an API call to get the latest release information
response = requests.get("https://api.github.com/repos/ami-iit/jaxsim/releases/latest")
data = response.json()

# Get the version from the response
version = data["tag_name"].replace("v", "")
```
In this way `readthedocs` will read the version directly from GitHub

diegoferigo:
> What about doing something like this in `conf.py`?:
> 
> ```python
> import requests
> 
> # Make an API call to get the latest release information
> response = requests.get("https://api.github.com/repos/ami-iit/jaxsim/releases/latest")
> data = response.json()
> 
> # Get the version from the response
> version = data["tag_name"].replace("v", "")
> ```
> 
> In this way `readthedocs` will read the version directly from GitHub and nothing would be hardcoded

Mmh I don't like this either, readthedocs can also render documentation from branches and old releases. The version should be directly inferred either from the sources or the corresponding python package.

I like https://github.com/ami-iit/jaxsim/pull/67#issuecomment-1907854332 since it always expose `jaxsim.__version__`, that might be useful downstream. For other packages, I occasionally used this variable for different reasons. Can you confirm that this is enough for readthedocs?

flferretti:
It should be enough, I tried it locally

PR #71, opened 2024-01-31
Update `jaxsim.typing` module

diegoferigo:
The current `jaxsim.typing` module is rather old, it was developed when `jax.Array` and `jax.typing.ArrayLike` did not exist.

Now that we have them, we can transition towards something more readable. I propose the following target:

- Use `jaxsim.typing.*Like` for type hinting input arguments that might be python scalars, numpy arrays, or jax arrays (note that static arguments must be hashable and therefore cannot be arrays).
- Use `jaxsim.typing.*` (without `Like`) for type hinting output arguments, that are all jax numpy arrays.

We currently have `jaxsim.typing.*Jax` for jax-only types, that should be ported to their corresponding types without the `Jax` suffix.

----
📚 Documentation preview 📚: https://jaxsim--71.org.readthedocs.build/en/71/

### DIFF
--- a/src/jaxsim/typing.py
+++ b/src/jaxsim/typing.py
@@ -1,39 +1,37 @@
-from typing import Any, Dict, Hashable, List, NamedTuple, Tuple, Union
+from typing import Any, Hashable
 
-import jax.numpy as jnp
-import numpy.typing as npt
+import jax
 
+# =========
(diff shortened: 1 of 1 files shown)


flferretti (approved):
LGTM

PR #74, opened 2024-02-01
Fix ReadTheDocs preview url in PRs

diegoferigo:
Solves preview url problem described in https://github.com/ami-iit/jaxsim/pull/73#issuecomment-1921594068.

----
📚 Documentation preview 📚: https://jaxsim--74.org.readthedocs.build/en/74/

### DIFF
--- a/.github/workflows/read_the_docs.yml
+++ b/.github/workflows/read_the_docs.yml
@@ -15,3 +15,4 @@ jobs:
       - uses: readthedocs/actions/preview@v1
         with:
           project-slug: "jaxsim"
+          project-language: ""


diegoferigo:
I'm not sure that the empty string is effective. This PR still uses the old workflow. I guess we'll figure this out in the next PR after merging this one.

flferretti:
Thanks @diegoferigo! You can also use [`act`](https://github.com/nektos/act) to test the workflows files locally

flferretti (approved):
LGTM, let's see if that works

diegoferigo:
> Thanks @diegoferigo! You can also use [`act`](https://github.com/nektos/act) to test the workflows files locally

I know what it is but I never used it yet. Do you think that it's useful also in this case, where the trigger is opening a PR and the result is an update of the first message?

For the time being, I'll be merging as it is. I'll fix the behavior in a new PR, if needed.

PR #77, opened 2024-02-02
Use structural pattern matching for `jcal`

flferretti:
This PR focuses on optimizing the computation speed of core algorithms modules that rely on `jcalc`, e.g. ABA, CRBA, RNEA and soft contacts. The implementation introduces [structural pattern matching](https://peps.python.org/pep-0622/), to replace extensive `if-else` statements. The anticipated outcome is an average speed improvement of about 80% in functions handling joint motion subspace or joint transformation.

`if-else` implementation:
```python
%timeit -n 5 _ = model.physics_model.joint_transforms(q)
>>> 884 ms ± 34.4 ms per loop (mean ± std. dev. of 7 runs, 5 loops each)

%timeit -n 5 _ = model.physics_model.motion_subspaces(q)
>>> 879 ms ± 28.2 ms per loop (mean ± std. dev. of 7 runs, 5 loops each)
```

`structural pattern matching` implementation:
```python
%timeit -n 5 _ = model.physics_model.motion_subspaces(q)
>>> 173 ms ± 1.28 ms per loop (mean ± std. dev. of 7 runs, 5 loops each)

%timeit -n 5 _ = model.physics_model.joint_transforms(q)
>>> 178 ms ± 6.12 ms per loop (mean ± std. dev. of 7 runs, 5 loops each)
```

----
📚 Documentation preview 📚: https://jaxsim--77.org.readthedocs.build//77/

### DIFF
--- a/src/jaxsim/math/joint.py
+++ b/src/jaxsim/math/joint.py
@@ -32,70 +32,71 @@ def jcalc(
     else:
         raise ValueError(jtyp)
 
-    if code is JointType.F:
-        raise ValueError("Fixed joints shouldn't be here")
+    match code:
+        case JointType.F:
+            raise ValueError("Fixed joints shouldn't be here")
 
-    if code is JointType.R:
-        jtyp: JointGenericAxis
+        case JointType.R:
+            jtyp: JointGenericAxis
 
-        Xj = Adjoint.from_rotation_and_translation(
-            rotation=Rotation.from_axis_angle(vector=q * jtyp.axis), inverse=True
-        )
+            Xj = Adjoint.from_rotation_and_translation(
+                rotation=Rotation.from_axis_angle(vector=q * jtyp.axis), inverse=True
+            )
 
-        S = jnp.vstack(jnp.hstack([jnp.zeros(3), jtyp.axis.squeeze()]))
+            S = jnp.vstack(jnp.hstack([jnp.zeros(3), jtyp.axis.squeeze()]))
 
-    elif code is JointType.P:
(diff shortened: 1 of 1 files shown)


diegoferigo (approved):
Cool, thanks for the improvement! It would be interesting to understand the performance of the jitted execution since we never run the vanilla code. Maybe once lowered to XLA, the speed becomes comparable.

Few more details:

- The joint type is a static member, stored in [`PhysicsModel._jtype_dict`](https://github.com/ami-iit/jaxsim/blob/ed05df47a0484280d478f33712e34a69ccb67d4c/src/jaxsim/physics/model/physics_model.py#L53-L55). Therefore, once compiled, only one branch becomes part of the compiled code. I suspect that this is similarly the case also with the `match-case` statement.
- Regardless, using `match-case` is more readable then `if-else`. The `jcalc` function was developed before Python 3.10.
- Since [JointType](https://github.com/ami-iit/jaxsim/blob/ed05df47a0484280d478f33712e34a69ccb67d4c/src/jaxsim/parsers/descriptions/joint.py#L14) is an `IntEnum` using auto, the values of its fields are integers starting from 0. Probably we can rewrite the `jcalc` function using `jax.lax.switch`, that would maybe allow us to make the joint type not static. I have no clue how the the speed of both compilation and runtime would be affected (either positively or negatively). To do that, we need an enum class compatible with jax, I think equinox has one.

flferretti:
Thanks Diego! For the third point I guess we can open an issue to discuss about it

PR #72, opened 2024-01-31
Add adaptive integrators (variable step)

diegoferigo:
----
📚 Documentation preview 📚: https://jaxsim--72.org.readthedocs.build/72/

PR #78, opened 2024-02-02
Enhance maintainability and performance

flferretti:
This PR aims to improve code readability, efficiency, and maintainability. In particular:

- Introduces a registry for managing calls to integrators, improving overall readability and maintainability
- Increases memory efficiency in [`high_level.Model.valid()`](https://github.com/ami-iit/jaxsim/blob/0cdfa794ef58adaeea617420a9d7d1008c7311b2/src/jaxsim/high_level/model.py#L320-L326) with generators 
- Improves readability in dictionaries iterations
- Removes calls to `dict()` for better performance (see [Performance Analysis of Python’s dict() and {}](https://madebyme.today/blog/python-dict-vs-curly-brackets/))  
- Implements set comprehension instead of using list comprehension and the convert them to sets
- Cleans up the codebase by removing unnecessary imports.

----
📚 Documentation preview 📚: https://jaxsim--78.org.readthedocs.build//78/

### DIFF
--- a/src/jaxsim/high_level/model.py
+++ b/src/jaxsim/high_level/model.py
@@ -321,8 +321,8 @@ class Model(Vmappable):
         """"""
 
         valid = True
-        valid = valid and all([l.valid() for l in self.links()])
-        valid = valid and all([j.valid() for j in self.joints()])
+        valid = valid and all(l.valid() for l in self.links())
+        valid = valid and all(j.valid() for j in self.joints())
         return jnp.array(valid, dtype=bool)
 
     @functools.partial(oop.jax_tf.method_ro, jit=False)
--- a/src/jaxsim/high_level/model.py
+++ b/src/jaxsim/high_level/model.py
@@ -1417,7 +1417,7 @@ class Model(Vmappable):
         assert isinstance(integrator_type, IntegratorType)
 
         # Integrate the model dynamics
-        ode_states, aux = ode_integration._integrator_registry[integrator_type](
+        ode_states, aux = ode_integration.ode_integration_fixed_step(
             x0=x0,
             t=jnp.array([t0, tf], dtype=float),
             ode_input=ode_input,
@@ -1425,6 +1425,7 @@ class Model(Vmappable):
             soft_contacts_params=contact_parameters,
             num_sub_steps=sub_steps,
             terrain=terrain,
+            integrator_type=integrator_type,
             return_aux=True,
         )
 
--- a/src/jaxsim/high_level/model.py
+++ b/src/jaxsim/high_level/model.py
@@ -1414,20 +1414,10 @@ class Model(Vmappable):
             physics_model=self.data.model_input
         )
 
-        if integrator_type is IntegratorType.EulerForward:
-            integrator_fn = ode_integration.ode_integration_euler
-
-        elif integrator_type is IntegratorType.EulerSemiImplicit:
-            integrator_fn = ode_integration.ode_integration_euler_semi_implicit
-
-        elif integrator_type is IntegratorType.RungeKutta4:
-            integrator_fn = ode_integration.ode_integration_rk4
-
-        else:
-            raise ValueError(integrator_type)
+        assert isinstance(integrator_type, IntegratorType)
 
         # Integrate the model dynamics
-        ode_states, aux = integrator_fn(
+        ode_states, aux = ode_integration._integrator_registry[integrator_type](
             x0=x0,
             t=jnp.array([t0, tf], dtype=float),
             ode_input=ode_input,
--- a/src/jaxsim/simulation/ode_integration.py
+++ b/src/jaxsim/simulation/ode_integration.py
@@ -19,6 +19,14 @@ class IntegratorType(enum.IntEnum):
     EulerSemiImplicitManifold = enum.auto()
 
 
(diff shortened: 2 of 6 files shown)


src/jaxsim/simulation/ode_integration.py
```diff
@@ -34,117 +42,25 @@ def ode_integration_rk4_adaptive(
...
-    )
-
-    # Return output pytree and, optionally, the aux dict
-    state = out if not return_aux else out[0]
-    return (state, out[1]) if return_aux else state
-
-
-@functools.partial(jax.jit, static_argnames=["num_sub_steps", "return_aux"])
-def ode_integration_euler_semi_implicit(
+def ode_integration(
```
diegoferigo, line 45:
```suggestion
def ode_integration_fixed_step(
```

src/jaxsim/simulation/ode_integration.py
```diff
@@ -34,117 +42,25 @@ def ode_integration_rk4_adaptive(
 
 
 @functools.partial(jax.jit, static_argnames=["num_sub_steps", "return_aux"])
```
diegoferigo, line 44:
```suggestion
@functools.partial(
    jax.jit, static_argnames=["num_sub_steps", "integrator_type", "return_aux"]
)
```

diegoferigo (approved):
Looks good, thanks! It makes sense to unify all the `ode_integration_*` functions since these integrators evolved to have the same APIs (compatible with [`jax.experimental.ode`](https://github.com/google/jax/blob/jaxlib-v0.4.23/jax/experimental/ode.py) by the way).  It can be done similarly to the [`odeint_*`](https://github.com/ami-iit/jaxsim/blob/72e130abd1fee0644feccfbe12d2e51f73891aa2/src/jaxsim/simulation/integrators.py#L335-L452) functions, but I'm considering to refactor soon this file with the new resources from #72, so let's keep them as they are - at least for now.

For the rest, just minor changes.

A comment. I'm not really sure if the performance analysis that usually apply to plain Python code is valid also here with jitted code. For example, one way to check e.g. the `dict` construction is to create two simple functions and check the resulting IR with [`jax.make_jaxpr`](https://jax.readthedocs.io/en/latest/_autosummary/jax.make_jaxpr.html). I suspect that it outputs same expression regardless to the construction. However, perhaps it runs more efficiently during the creation of the expression, so it may not be entirely useless.

src/jaxsim/simulation/ode_integration.py
flferretti: Good catch! Thanks

src/jaxsim/high_level/model.py
```diff
@@ -1414,20 +1414,10 @@ def integrate(
...
-        elif integrator_type is IntegratorType.RungeKutta4:
-            integrator_fn = ode_integration.ode_integration_rk4
-
-        else:
-            raise ValueError(integrator_type)
+        assert isinstance(integrator_type, IntegratorType)
 
         # Integrate the model dynamics
-        ode_states, aux = integrator_fn(
+        ode_states, aux = ode_integration._integrator_registry[integrator_type](
```
diegoferigo, line 1420: @flferretti I missed this in the review, I suspect you have to call the new generic `ode_fixed_step`, maybe creating a lambda.

flferretti: Yes, I was checking it right now. I'm going to do some tests and get back to you asap

flferretti: Fixed in https://github.com/ami-iit/jaxsim/pull/78/commits/73f8ca6d53f526730f07ded77ec2b521c3427390

flferretti:
> A comment. I'm not really sure if the performance analysis that usually apply to plain Python code is valid also here with jitted code. For example, one way to check e.g. the `dict` construction is to create two simple functions and check the resulting IR with [`jax.make_jaxpr`](https://jax.readthedocs.io/en/latest/_autosummary/jax.make_jaxpr.html). I suspect that it outputs same expression regardless to the construction. However, perhaps it runs more efficiently during the creation of the expression, so it may not be entirely useless.

You were right, the `jaxpr` is the same:

```python
In [3]: jax.make_jaxpr(lambda k,v: dict(k=v))(k,v)
Out[3]: { lambda ; a:i32[] b:i32[]. let  in (b,) }

In [4]: jax.make_jaxpr(lambda k,v: {k:v})(k,v)
Out[4]: { lambda ; a:i32[] b:i32[]. let  in (b,) }
```

It does run slightly faster, but I'm pretty sure it won't make a huge difference when the code is compiled

```python
In [3]: %timeit dict(k=v)
63.3 ns ± 4.07 ns per loop (mean ± std. dev. of 7 runs, 10,000,000 loops each)

In [4]: %timeit {k:v}
52.8 ns ± 3.91 ns per loop (mean ± std. dev. of 7 runs, 10,000,000 loops each)

In [5]: %timeit jax.make_jaxpr(lambda k,v: dict(k=v))(k,v)
186 µs ± 6.1 µs per loop (mean ± std. dev. of 7 runs, 10,000 loops each)

In [6]: %timeit jax.make_jaxpr(lambda k,v: {k:v})(k,v): 
179 µs ± 2.36 µs per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
```

PR #80, opened 2024-02-05
Run `pytest` in CI only in PRs that update either sources or tests

diegoferigo:
Since running `pytest` is pretty slow (~1h), let's mitigate our contribution on global warming by preventing to run the corresponding CI step when not needed.

I used https://github.com/dorny/paths-filter that seems quite popular.

Note that this change only targets PRs, not CI pipelines on branches. We just need green CI on PRs for merging.

----
📚 Documentation preview 📚: https://jaxsim--80.org.readthedocs.build//80/

### DIFF
--- a/.github/workflows/ci_cd.yml
+++ b/.github/workflows/ci_cd.yml
@@ -65,7 +65,7 @@ jobs:
       - name: Set up Python
         uses: actions/setup-python@v4
         with:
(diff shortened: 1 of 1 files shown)


diegoferigo:
I'll squash merge this PR, ignore the dirty commit names.

PR #81, opened 2024-02-07
Add GitHub Codespaces support

diegoferigo:
This PR adds the bare minimum resources to obtain a GitHub Codespace with the following features:

- Docker container based on Ubuntu.
- Python environment with all the JAXsim dependencies installed in a conda environment.
- Capable of navigating the Python project.
- Support of autocompletion.

This should help the process of on-the-fly editing and reviewing.

---

Here below further details:

- The base image in which VSCode runs is the latest LTS as documented [here](https://github.com/devcontainers/images/tree/main/src/base-ubuntu).
- The initial `devcontainer.json` was taken by [this template](https://github.com/devcontainers/templates/tree/main/src/ubuntu).
- The conda environment is created by enabling the [micromamba](https://github.com/mamba-org/devcontainer-features/tree/main/src/micromamba) feature (you can find [here](https://containers.dev/features) the complete list of features). It requires extending the base image if the environment is created from a yaml file as in our case.
- I used and slightly modified the yaml file originally added by @flferretti for ReadTheDocs, moving it to the top-level. As a consequence, the repo now supports other web tools that search for top-level environment files (like [binder](https://github.com/binder-examples/conda)).
- I enabled, for now, only the [official Python extension](https://marketplace.visualstudio.com/items?itemName=ms-python.python).

The steps to open a codespace is the following:

1. Open the desired branch in the github repository.
1. Click the green button `Code > Create workspace in devcontainer`.
1. The image is built and the container is started.
1. As soon as VSCode shows up, the terminal in the bottom should already have the `jaxsim` conda environment generated from the yaml file ready and active (check `which python`).
1. Wait a few seconds more to let VSCode finish the installation of the active extensions.
1. Open any Python file, and in the bottom-right corner select the Python 3.11 interpreter. By default, it detects the system interpreter.

Now the project is ready to be navigated. It is not necessary to install the project in editable mode if no code needs to be executed. VSCode automatically detects the sources in `src/` and Ctrl+click should work fine. Otherwise, to e.g. run any code (tests included), the project can be installed as follows:

```bash
pip install --no-deps -e .
``` 

---

Other notes:

- I had to be creative to enable the conda environment and make its Python interpreter show up in the list. I have a custom [dotfiles](https://github.com/diegoferigo/dotfiles) repo that initializes my shell, and I fear that it interferes with the `~/.bashrc` initialization done by the micromamba feature. Regardless, I managed to work around the problems as explained.
- I tried to pip-install the project in the `postCreateCommand` of `devcontainer.json` but either it doesn't run in shell that is initialized with the conda environment, or it runs too early when the environment is not ready. If the problem is the first one, probably we can do it by defining a container `entrypoint.sh`.
- Introduction to devcontainers: https://aka.ms/codespaces-customize.
- Documentation of `devcontainer.json`: https://containers.dev/implementors/json_reference/.
- List of supported features: https://containers.dev/features.
- How to enable a feature: https://docs.github.com/en/codespaces/setting-up-your-project-for-codespaces/configuring-dev-containers/adding-features-to-a-devcontainer-file.

Running the tests now fail due to:

```
ImportError while importing test module '/workspaces/jaxsim/tests/test_eom.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
/opt/conda/envs/jaxsim/lib/python3.11/importlib/__init__.py:126: in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
tests/test_eom.py:11: in <module>
    from . import utils_idyntree, utils_models, utils_rng
tests/utils_idyntree.py:5: in <module>
    import idyntree.bindings as idt
/opt/conda/envs/jaxsim/lib/python3.11/site-packages/idyntree/__init__.py:1: in <module>
    from . import swig
/opt/conda/envs/jaxsim/lib/python3.11/site-packages/idyntree/swig.py:10: in <module>
    from . import _iDynTree
E   ImportError: libGL.so.1: cannot open shared object file: No such file or directory
```

I'm not interested to fix this at this stage. cc FYI @traversaro, likely I need to install some missing system dependency or GL support in the conda environment.

----
📚 Documentation preview 📚: https://jaxsim--81.org.readthedocs.build//81/


traversaro:
> FYI @traversaro, likely I need to install some missing system dependency or GL support in the conda environment.

Yes, this is due to https://conda-forge.org/docs/maintainer/knowledge_base.html#libgl . Unfortunatly there there is no docs on which apt packages should be installed, but via apt-file:

~~~
traversaro@IITBMP014LW012:~$ apt-file search libGL.so.1
libgl1: /usr/lib/x86_64-linux-gnu/libGL.so.1
libgl1: /usr/lib/x86_64-linux-gnu/libGL.so.1.7.0
nsight-compute: /usr/lib/nsight-compute/host/linux-desktop-glibc_2_11_3-x64/Mesa/libGL.so.1
nsight-compute: /usr/lib/nsight-compute/host/linux-desktop-glibc_2_11_3-x64/Mesa/libGL.so.1.5.0
nsight-systems: /usr/lib/nsight-systems/host-linux-x64/Mesa/libGL.so.1
nsight-systems: /usr/lib/nsight-systems/host-linux-x64/Mesa/libGL.so.1.5.0
primus-libs: /usr/lib/x86_64-linux-gnu/primus/libGL.so.1
~~~

so I guess `sudo apt install libgl1` should be ok.

flferretti (approved):
LGTM, thanks

traversaro (approved):

traversaro:
Thanks! For reference I was worried about possible billing for the ami-iit organization, but after reading https://docs.github.com/en/codespaces/managing-codespaces-for-your-organization/choosing-who-owns-and-pays-for-codespaces-in-your-organization we should be good to go (we have a 0€ limit for Codespaces).

diegoferigo:
> so I guess sudo apt install `libgl1` should be ok

Yep I can confirm that `libgl1` is the bare minimum required.

> I was worried about possible billing for the ami-iit organization

Good point, I didn't think of billings. How does it work, do we have some free hours included? I just checked my personal account since I was originally thinking that it would have consumed my free hours, but it seems that I didn't consume any.

Screenshot_20240207_221201

traversaro:
> > 
> 
> Good point, I didn't think of billings. How does it work, do we have some free hours included?

I don't think so, there is a " A $0 limit will prevent your organization from using Codespaces " notice under the limit.

PR #85, opened 2024-02-14
Add static rendering in `MujocoVideoRecorder`

flferretti:
This PR will add a method in the `MujocoVideoRecorder` class to return images to be rendered in Python Notebooks.

----
📚 Documentation preview 📚: https://jaxsim--85.org.readthedocs.build//85/

### DIFF
--- a/src/jaxsim/mujoco/visualizer.py
+++ b/src/jaxsim/mujoco/visualizer.py
@@ -52,13 +52,19 @@ class MujocoVideoRecorder:
         self.data = data if data is not None else self.data
         self.model = model if model is not None else self.model
 
-    def render_frame(self, camera_name: str | None = None) -> None:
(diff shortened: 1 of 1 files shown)


diegoferigo:
The logic of the new method you drafted is already in `MujocoVideoRecorder.render_frame`, but in that case we do not return the rendered frame and instead append it to the stored frames that will be converted to a video.

What about unifying that logic? I mean:

- Modifying the following method to return a frame without appending to `self.frames`:
  ```python
  def render_frame(self, camera_name: str | None = None) -> npt.NDArray:
  ```
- Adding a new method for rendering and appending:
  ```python
  def record_frame(self, camera_name: str | None = None) -> None:
      frame = self.render_frame(camera_name=camera_name)
      self.frames.append(frame)
  ```

diegoferigo (approved):
Awesome, thanks!

PR #88, opened 2024-02-20
Introduce new functional APIs

diegoferigo:
These new functional APIs (`jaxsim.api`) will soon replace those based on OOP (`jaxsim.high_level`).

- Switch to [kiss](https://en.wikipedia.org/wiki/KISS_principle). The former OOP pattern, although more user friendly, fights back the functional JAX approach and required complex decorators to maintain and debug.
- Easier for users to extend / modify the logic. Instead of changing something deep down in the framework, they just have to update / rewrite a single function.
- Applying jax transforms like `jax.vmap` and `jax.grad` is more simple being more vanilla.

Furthermore, this PR:

- Introduces new functions to operate on links and joints through their indices. Although I never liked this approach (and using names is still an option), having also functions operating on indices enables the use of `jax.vmap` to iterate on links and joints.
- Introduces a new logic to get a good initial set of soft-contacts parameters based on the model inertial properties. In the past, tuning these parameters by hand has been quite difficult. The new logic provides a good starting point from which users can start iterating.
- The existing fixed-step integrators have been rewritten from scratch. The new structure is now compatible with the introduction of variable-step integrators (in a future PR). So far, only the explicit schemes have been ported. This time, all of them with variants that integrate the quaternion on $\text{SO}(3)$.

Some implementation detail that might ease the review process and the transition to the new APIs:

- The key data structures are now `JaxSimModel` and `JaxSimModelData`.
- Most of API functions take them as first two arguments. In order to have more stable APIs, and allow us to change the signature of functions in future versions, all others arguments are enforced to be `kwargs`.
- The velocity representation is now part of data (still as static attribute).
- The soft-contacts parameters is now part of data. Therefore, in a vectorized scenario, there could be different parameters for each data instance, enabling domain randomization on these parameters.
- The terrain instead is global and stored inside the model (therefore, it is not possible to parallelize a model over different terrains, that makes sense at least for now).

Next steps:

- Port semi-implicit integration schemes to the new APIs.
- Introduce variable-step integrators.
- Merge `jaxsim.api.model.JaxSimModel` with `jaxsim.physics.model.PhysicsModel`.
- Introduce a new pytree `KynDynParameters` to store inside `JaxSimModel` the model parameters, so that we can differentiate against them as well (we need also decent setters/getters).
- Introduce error handling for link/joint indices out of bound, and data objects not compatible with model.

----
📚 Documentation preview 📚: https://jaxsim--88.org.readthedocs.build//88/

### DIFF
--- a/src/jaxsim/api/data.py
+++ b/src/jaxsim/api/data.py
@@ -641,7 +641,7 @@ def random_model_data(
         ).as_quaternion_xyzw()[np.array([3, 0, 1, 2])]
 
         physics_model_state.joint_positions = jaxsim.api.joint.random_joint_positions(
-            model=model
+            model=model, key=k3
         )
 
         physics_model_state.base_linear_velocity = jax.random.uniform(
--- a/src/jaxsim/api/data.py
+++ b/src/jaxsim/api/data.py
@@ -259,6 +259,7 @@ class JaxSimModelData(JaxsimDataclass):
 
         return self.time_ns.astype(float) / 1e9
 
+    @functools.partial(jax.jit, static_argnames=["joint_names"])
     def joint_positions(
         self,
         model: jaxsim.api.model.JaxSimModel | None = None,
@@ -298,6 +299,7 @@ class JaxSimModelData(JaxsimDataclass):
             jaxsim.api.joint.names_to_idxs(joint_names=joint_names, model=model)
         ]
 
+    @functools.partial(jax.jit, static_argnames=["joint_names"])
     def joint_velocities(
         self,
         model: jaxsim.api.model.JaxSimModel | None = None,
--- a/src/jaxsim/api/data.py
+++ b/src/jaxsim/api/data.py
@@ -246,6 +246,20 @@ class JaxSimModelData(JaxsimDataclass):
             velocity_representation=velocity_representation,
         )
 
+    # ==================
+    # Extract quantities
+    # ==================
+
+    def time(self) -> jtp.Float:
+        """
+        Get the simulated time.
+
+        Returns:
+            The simulated time in seconds.
+        """
+
+        return self.time_ns.astype(float) / 1e9
+
     def joint_positions(
         self,
         model: jaxsim.api.model.JaxSimModel | None = None,
--- a/src/jaxsim/api/data.py
+++ b/src/jaxsim/api/data.py
@@ -20,7 +20,6 @@ import jaxsim.physics.algos.rnea
 import jaxsim.physics.model.physics_model
 import jaxsim.physics.model.physics_model_state
 import jaxsim.typing as jtp
-from jaxsim import sixd
 from jaxsim.high_level.common import VelRepr
 from jaxsim.physics.algos import soft_contacts
 from jaxsim.simulation.ode_data import ODEState
@@ -218,13 +217,13 @@ class JaxSimModelData(JaxsimDataclass):
         v_WB = JaxSimModelData.other_representation_to_inertial(
             array=jnp.hstack([base_linear_velocity, base_angular_velocity]),
             other_representation=velocity_representation,
-            base_transform=W_H_B,
+            transform=W_H_B,
             is_force=False,
         )
 
         ode_state = ODEState.build(
             physics_model=model.physics_model,
-            physics_model_state=jaxsim.physics.model.physics_model.PhysicsModelState(
+            physics_model_state=jaxsim.physics.model.physics_model_state.PhysicsModelState(
                 base_position=base_position.astype(float),
                 base_quaternion=base_quaternion.astype(float),
                 joint_positions=joint_positions.astype(float),
@@ -377,7 +376,7 @@ class JaxSimModelData(JaxsimDataclass):
         return (
             base_unit_quaternion
             if not dcm
-            else sixd.so3.SO3.from_quaternion_xyzw(
+            else jaxlie.SO3.from_quaternion_xyzw(
                 base_unit_quaternion[to_xyzw]
             ).as_matrix()
         )
@@ -423,7 +422,7 @@ class JaxSimModelData(JaxsimDataclass):
             JaxSimModelData.inertial_to_other_representation(
                 array=W_v_WB,
                 other_representation=self.velocity_representation,
-                base_transform=W_H_B,
+                transform=W_H_B,
                 is_force=False,
             )
             .squeeze()
@@ -712,7 +711,7 @@ class JaxSimModelData(JaxsimDataclass):
         W_v_WB = self.other_representation_to_inertial(
             array=jnp.atleast_1d(base_velocity.squeeze()).astype(float),
             other_representation=velocity_representation,
-            base_transform=self.base_transform(),
+            transform=self.base_transform(),
             is_force=False,
         )
 
@@ -735,7 +734,7 @@ class JaxSimModelData(JaxsimDataclass):
     def inertial_to_other_representation(
         array: jtp.Array,
         other_representation: VelRepr,
-        base_transform: jtp.Matrix,
+        transform: jtp.Matrix,
         is_force: bool = False,
     ) -> jtp.Array:
         """
@@ -744,7 +743,9 @@ class JaxSimModelData(JaxsimDataclass):
         Args:
             array: The 6D quantity to convert.
             other_representation: The representation to convert to.
-            base_transform: The base transform.
+            transform:
+                The `math:W \mathbf{H}_O` transform, where `math:O` is the
+                reference frame of the other representation.
             is_force: Whether the quantity is a 6D force or 6D velocity.
 
         Returns:
@@ -752,13 +753,13 @@ class JaxSimModelData(JaxsimDataclass):
         """
 
         W_array = array.squeeze()
-        W_H_B = base_transform.squeeze()
+        W_H_O = transform.squeeze()
 
         if W_array.size != 6:
             raise ValueError(W_array.size, 6)
 
-        if W_H_B.shape != (4, 4):
-            raise ValueError(W_H_B.shape, (4, 4))
+        if W_H_O.shape != (4, 4):
+            raise ValueError(W_H_O.shape, (4, 4))
 
         match other_representation:
 
@@ -768,28 +769,28 @@ class JaxSimModelData(JaxsimDataclass):
             case VelRepr.Body:
 
                 if not is_force:
-                    B_Xv_W = sixd.se3.SE3.from_matrix(W_H_B).inverse().adjoint()
-                    B_array = B_Xv_W @ W_array
+                    O_Xv_W = jaxlie.SE3.from_matrix(W_H_O).inverse().adjoint()
+                    O_array = O_Xv_W @ W_array
 
                 else:
-                    B_Xf_W = sixd.se3.SE3.from_matrix(W_H_B).adjoint().T
-                    B_array = B_Xf_W @ W_array
+                    O_Xf_W = jaxlie.SE3.from_matrix(W_H_O).adjoint().T
+                    O_array = O_Xf_W @ W_array
 
-                return B_array
+                return O_array
 
             case VelRepr.Mixed:
-                W_p_B = W_H_B[0:3, 3]
-                W_H_BW = jnp.eye(4).at[0:3, 3].set(W_p_B)
+                W_p_O = W_H_O[0:3, 3]
+                W_H_OW = jnp.eye(4).at[0:3, 3].set(W_p_O)
 
                 if not is_force:
-                    BW_Xv_W = sixd.se3.SE3.from_matrix(W_H_BW).inverse().adjoint()
-                    BW_array = BW_Xv_W @ W_array
+                    OW_Xv_W = jaxlie.SE3.from_matrix(W_H_OW).inverse().adjoint()
+                    OW_array = OW_Xv_W @ W_array
 
                 else:
-                    BW_Xf_W = sixd.se3.SE3.from_matrix(W_H_BW).adjoint().T
-                    BW_array = BW_Xf_W @ W_array
+                    OW_Xf_W = jaxlie.SE3.from_matrix(W_H_OW).adjoint().transpose()
+                    OW_array = OW_Xf_W @ W_array
 
-                return BW_array
+                return OW_array
 
             case _:
                 raise ValueError(other_representation)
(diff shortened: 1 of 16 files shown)


src/jaxsim/integrators/common.py
```diff
@@ -0,0 +1,416 @@
...
+        rows_of_A_with_fsal = jnp.logical_and(rows_of_A_with_fsal, (c == 1))
+
+        # If there is no match, it means that the Butcher tableau does not support FSAL.
+        if not rows_of_A_with_fsal.any():
+            return False, None
+
+        # Return the index of the row of A providing the fsal derivative (that is the
+        # possibly intermediate kᵢ derivative).
+        # Note that if multiple rows match (it should not), we return the first match.
+        return True, int(jnp.where(rows_of_A_with_fsal == True)[0].tolist()[0])
```
flferretti, line 416: This might create issues as it is an implicit boolean conversion of a traced array

```suggestion
        return True, int(jnp.where(rows_of_A_with_fsal is True)[0].tolist()[0])
```

src/jaxsim/api/model.py
flferretti, line 1: `functools` and `VelRep` are not imported

src/jaxsim/integrators/common.py
diegoferigo: Generally speaking this would be true, but this method is not meant to run in any jitted context. I believe that in that case, the logic would fail first since the output of the method is not constant (either `None` or `int`).

The integrators are meant to be pytrees without leafs (for the moment) that are built outside transformed functions. Note, however, that `Integrator.step` could be jitted and vmapped.

src/jaxsim/physics/model/physics_model.py
```diff
@@ -70,11 +70,47 @@ class PhysicsModel(JaxsimDataclass):
...
+            )
+        ]
+
+        ordered_joints = [
+            j
+            for j in sorted(
+                list(self.description.joints_dict.values()),
+                key=lambda j: j.index,
+            )
+        ]
```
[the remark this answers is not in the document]
diegoferigo: Yep thanks for the suggestion, it makes sense. Let me leave this suggestion unresolved for now. This logic is needed to initialize new attributes of `PhysicsModel` that are still in WIP, and they are the candidates to be externalized in a new `KinDynParameters` class that might allow us to obtain a parametric model. In any case, I will likely include this new feature in a new PR.

diegoferigo:
The MWE of the new functional APIs is the following script. Resulting video attached.

<details>
<summary>MWE</summary>

```python
import pathlib

import jax.numpy as jnp
import jaxsim.api as js
import numpy as np
import rod
from jaxsim import VelRepr, integrators

# Find the urdf file.
)

# Build the ROD model.
rod_sdf = rod.Sdf.load(sdf=urdf_path)

# Build the model.
    model_description=rod_sdf.model,
    gravity=jnp.array([0, 0, -10.0]),
)

# Reduce the model.
model = js.model.reduce(
    model=model,
    considered_joints=tuple(
        [
            j
            for j in model.joint_names()
            if "camera" not in j
            and "neck" not in j
            and "wrist" not in j
            and "thumb" not in j
            and "index" not in j
            and "middle" not in j
            and "ring" not in j
            and "pinkie" not in j
        ]
    ),
)

# Build the model's data.
# Set already here the initial base position.
data0 = js.data.JaxSimModelData.build(
    model=model,
    base_position=jnp.array([0, 0, 0.85]),
    velocity_representation=VelRepr.Inertial,
)

# Update the soft-contact parameters.
# By default, only 1 support point is used as worst-case scenario.
# Feel free to tune this with more points to get a less stiff system.
data0 = data0.replace(
    soft_contacts_params=js.contact.estimate_good_soft_contacts_parameters(
        model, number_of_active_collidable_points_steady_state=2
    )
)

# =====================
# Create the integrator
# =====================

# Create a RK4 integrator integrating the quaternion on SO(3).
integrator = integrators.fixed_step.RungeKuttaSO3.build(
    dynamics=js.ode.wrap_system_dynamics_for_integration(
        model=model,
        data=data0,
        system_dynamics=js.ode.system_dynamics,
    ),
)

# =========================================
# Visualization in Mujoco viewer / renderer
# =========================================

from jaxsim.mujoco import MujocoModelHelper, MujocoVideoRecorder, RodModelToMjcf

# Convert the ROD model to a Mujoco model.
mjcf_string, assets = RodModelToMjcf.convert(
    rod_model=rod_sdf.models()[0],
    considered_joints=list(model.joint_names()),
)

# Build the Mujoco model helper.
mj_model_helper = self = MujocoModelHelper.build_from_xml(
    mjcf_description=mjcf_string, assets=assets
)

# Create the video recorder.
recorder = MujocoVideoRecorder(
    model=mj_model_helper.model,
    data=mj_model_helper.data,
    fps=int(1 / 0.010),
    width=320 * 4,
    height=240 * 4,
)

# ==============
# Recording loop
# ==============

# Initialize the integrator.
t0 = 0.0
tf = 5.0
dt = 0.001_000
integrator_state = integrator.init(x0=data0.state, t0=t0, dt=dt)

# Initialize the loop.
data = data0.copy()
joint_names = list(model.joint_names())

while data.time_ns < tf * 1e9:

    # Integrate the dynamics.
    data, integrator_state = js.model.step(
        dt=dt,
        model=model,
        data=data,
        integrator=integrator,
        integrator_state=integrator_state,
        # Optional inputs
        joint_forces=None,
        external_forces=None,
    )

    # Extract the generalized position.
    s = data.state.physics_model.joint_positions
    W_p_B = data.state.physics_model.base_position
    W_Q_B = data.state.physics_model.base_quaternion

    # Update the data object stored in the helper, which is shared with the recorder.
    mj_model_helper.set_base_position(position=np.array(W_p_B))
    mj_model_helper.set_base_orientation(orientation=np.array(W_Q_B), dcm=False)
    mj_model_helper.set_joint_positions(positions=np.array(s), joint_names=joint_names)

    # Record the frame if the time is right to get the desired fps.
    if data.time_ns % jnp.array(1e9 / recorder.fps).astype(jnp.uint64) == 0:
        recorder.render_frame(camera_name=None)

# Store the video.
video_path = pathlib.Path("~/video.mp4").expanduser()
recorder.write_video(path=video_path, exist_ok=True)

# Clean up the recorder.
recorder.frames = []
recorder.renderer.close()
```

</details>

<details>
<summary>Video</summary>

https://github.com/ami-iit/jaxsim/assets/469199/d92e53e3-e71b-4b5a-84e4-e563d94dec3e

</details>

Some comment:

- With the automatic computation of the soft-contact parameters, using the fixed-step RK4 integrator with quaternion on $\text{SO}(3)$ with $\Delta t = 0.001 s$ is enough to have a stable simulation.
- Stepping the ergoCub robot, reduced to have 23 DoFs, takes:
  ```
  %timeit js.model.step(...)
  968 µs ± 137 µs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)
  ```
  obtaining therefore a real-time-factor of 1.

src/jaxsim/api/model.py
```diff
@@ -0,0 +1,1099 @@
...
+        A tuple containing the 6D acceleration in the active representation of the
+        base link and the joint accelerations resulting from the application of the
+        considered joint forces and external forces.
+    """
+
+    forward_dynamics_fn = forward_dynamics_aba if prefer_aba else forward_dynamics_crb
+
+    return forward_dynamics_fn(
+        model=model, data=data, tau=joint_forces, external_forces=external_forces
+    )
```
flferretti, line 485: `tau` is not present as a keyword argument in both ABA and CRB, typo?
```suggestion
    return forward_dynamics_fn(
        model=model, data=data, joint_forces=joint_forces, external_forces=external_forces
    )
```

flferretti:
Thanks for working on this @diegoferigo, it looks pretty neat! Could you please add the new modules in the docs?

src/jaxsim/physics/algos/soft_contacts.py
```diff
@@ -1,3 +1,5 @@
+from __future__ import annotations
+
```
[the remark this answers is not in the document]
diegoferigo: I need this every time the following pattern is used:

```python
class MyClass:
    @staticmethod
    def build() -> MyClass:
        pass
``` 

It allows to use `-> MyClass` instead of `-> "MyClass"`. I discovered it not long time ago :)

src/jaxsim/integrators/common.py
```diff
@@ -0,0 +1,503 @@
...
+        pass
+
+    def init(
+        self,
+        x0: State,
+        t0: Time,
+        dt: TimeStep,
+        *,
+        key: jax.Array | None = None,
+        **kwargs,
```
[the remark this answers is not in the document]
diegoferigo: It is not used.... now. It is there since users can create their own integrators. Maybe a random key is necessary, and here `kwargs` is used differently since it must match the `kwargs` of the `__call__` method.

src/jaxsim/api/link.py
```diff
@@ -0,0 +1,262 @@
...
+        model: The model to consider.
+        data: The data of the considered model.
+        link_index: The index of the link.
+
+    Returns:
+        The 4x4 matrix representing the transform.
+    """
+
+    return Model.forward_kinematics(model=model, data=data)[link_index]
+
```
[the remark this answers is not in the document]
diegoferigo: I guess this commit was done on an old version of this PR, I can no longer find any `jaxsim.helper` reference.

src/jaxsim/api/data.py
```diff
@@ -0,0 +1,663 @@
...
+        base_vel_lin_bounds: The bounds for the base linear velocity.
+        base_vel_ang_bounds: The bounds for the base angular velocity.
+        joint_vel_bounds: The bounds for the joint velocities.
+
+    Returns:
+        A `JaxSimModelData` object with random data.
+    """
+
+    key = key if key is not None else jax.random.PRNGKey(seed=0)
+    k1, k2, k3, k4, k5, k6 = jax.random.split(key, num=6)
```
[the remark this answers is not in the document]
diegoferigo: Good catch! Fixed in 8477d1b

src/jaxsim/api/data.py
```diff
@@ -0,0 +1,663 @@
...
+        Args:
+            joint_names:
+                The names of the joints for which to get the positions. If `None`, the
+                positions of all joints are returned.
+
+        Returns:
+            The joint positions.
+        """
+
+        import jaxsim.api.joint
```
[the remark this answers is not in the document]
diegoferigo: Fixed in https://github.com/ami-iit/jaxsim/pull/88/commits/1ae54bce5fee0a6569861685212a154ccb98834d

diegoferigo:
> Thanks for working on this @diegoferigo, it looks pretty neat! Could you please add the new modules in the docs?

I don't want to waste time now to fix the documentation, I'll do it as soon as things get more stable. I expect problem with math rendering etc, and I don't want to dedicate time to it at this early stage :) It has to be done before the next release.

DanielePucci:
> I don't want to dedicate time to it at this early stage :)

Agreed!

flferretti (approved):
Thanks a lot Diego, LGTM

src/jaxsim/integrators/common.py
```diff
@@ -0,0 +1,503 @@
+import abc
+import dataclasses
+from typing import Any, ClassVar, Generic, Protocol, Self, Type, TypeVar
```
flferretti, line 3: This ensures partial compatibility with Python 3.10, which is nice to have at this stage

```suggestion
from typing import Any, ClassVar, Generic, Protocol, Type, TypeVar

try:
    from typing import Self
except ImportError:
    from typing_extensions import Self
```

flferretti:
I have a couple more comments:
- If we want the old OOP structure to follow a deprecation cycle, I believe it can be worth adapting the tests in order to validate the new API. It's probably worth testing it anyway
- It's not clear to me how to change the pose parameters. In the older API you could call e.g.  `model.reset_joint_positions(...)`, what should a user modify now, the JaxsimModelData using the `replace` method?

C.C. @traversaro

src/jaxsim/integrators/common.py
diegoferigo: We ditched on purpose Python 3.10, in fact you can see:

https://github.com/ami-iit/jaxsim/blob/389c556de16308505d65ae9f38e7dd74108fdb4d/setup.cfg#L53

diegoferigo:
> I have a couple more comments:
> 
> * If we want the old OOP structure to follow a deprecation cycle, I believe it can be worth adapting the tests in order to validate the new API. It's probably worth testing it anyway

I already have local tests. They will be added in a new PR.

> * It's not clear to me how to change the pose parameters. In the older API you could call e.g.  `model.reset_joint_positions(...)`, what should a user modify now, the JaxsimModelData using the `replace` method? In that case, if I'm not mistaken, it would be something like:

Although the new functional APIs will get merged into `main`, they are not yet 1:1 with the OOP ones. I'm still working on resources to simplify reset and actuation. Now you need to alter directly `data.state` as you found out.

src/jaxsim/integrators/common.py
flferretti: Yes, but for some internal lab activities, I'm still using it with python 3.10 as some libraries are still not compatible with python 3.11. In addition, I've seen the same logic in other parts of the code: https://github.com/ami-iit/jaxsim/blob/389c556de16308505d65ae9f38e7dd74108fdb4d/tests/test_jax_oop.py#L14-L17, https://github.com/ami-iit/jaxsim/blob/389c556de16308505d65ae9f38e7dd74108fdb4d/src/jaxsim/utils/vmappable.py#L10-L13, https://github.com/ami-iit/jaxsim/blob/389c556de16308505d65ae9f38e7dd74108fdb4d/src/jaxsim/utils/jaxsim_dataclass.py#L14-L17, https://github.com/ami-iit/jaxsim/blob/389c556de16308505d65ae9f38e7dd74108fdb4d/src/jaxsim/simulation/simulator.py#L6-L9
wouldn't it be better to unify it everywhere or delete it everywhere?

diegoferigo: In this case, it makes sense to add -unofficial- Python 3.10 support. It's minor changes, for now. I expect that we migrate to Ubuntu 24.04 sooner or later, with Python 3.11 as default.

diegoferigo:
@flferretti I just added `JaxSimModelData.reset_*` methods.

src/jaxsim/integrators/common.py
diegoferigo: Done in https://github.com/ami-iit/jaxsim/pull/88/commits/4e49b171157c270b5c4d17bb04de5ba589b51232

PR #92, opened 2024-02-27
Initialize JAX configuration as per the v0.4.25 release

flferretti:
This PR will update the JAX configuration initialization as per [JAX v0.4.25](https://github.com/google/jax/releases/tag/jax-v0.4.25), in which the access to `jax.config` has been deprecated, leading to `ImportError: cannot import name 'config' from 'jax.config'`.

----
📚 Documentation preview 📚: https://jaxsim--92.org.readthedocs.build//92/

PR #97, opened 2024-03-01
Pin `jax<0.4.15` also in `environment.yml`

diegoferigo:
Otherwise readthedocs fails.


flferretti:
I think the actions on #88 fail because the branch is not update to #92. Is this actually needed?

environment.yml
```diff
@@ -5,8 +5,8 @@ dependencies:
   # Dependencies from setup.cfg
   - python=3.11
   - coloredlogs
-  - jax >= 0.4.13
-  - jaxlib
+  - jax >= 0.4.13,<0.4.15
+  - jaxlib < 0.4.15
```
flferretti, line 9: Shouldn't we put this as in [`setup.cfg`](https://github.com/ami-iit/jaxsim/blob/d0f4d4e0d1e0c0685b37e4f9fcac26c1466fb545/setup.cfg#L56C1-L57C30)? (jax<0.4.25):

```suggestion
  - jax >= 0.4.13,<0.4.25
  - jaxlib < 0.4.25
```

diegoferigo:
I haven't tried, let me check after rebasing. Since readthedocs is not executing any code, you might be right and the documentation could build successfully.

PR #99, opened 2024-03-06
Multiple bug fixing of new functional APIs

diegoferigo:
I'm in the process of developing unit tests of the new functional APIs and I found multiple bugs in velocity representations different from `VelRepr.Inertial`. This PR aims at fixing them.

### DIFF
--- a/src/jaxsim/api/data.py
+++ b/src/jaxsim/api/data.py
@@ -695,6 +695,7 @@ def random_model_data(
     model: jaxsim.api.model.JaxSimModel,
     *,
     key: jax.Array | None = None,
+    velocity_representation: VelRepr | None = None,
     base_pos_bounds: tuple[
         jtp.FloatLike | Sequence[jtp.FloatLike],
         jtp.FloatLike | Sequence[jtp.FloatLike],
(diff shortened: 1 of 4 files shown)


src/jaxsim/api/data.py
```diff
@@ -738,7 +740,14 @@ def random_model_data(
...
     ṡ_min, ṡ_max = joint_vel_bounds
 
-    random_data = JaxSimModelData.zero(model=model)
+    random_data = JaxSimModelData.zero(
+        model=model,
+        **(
+            dict(velocity_representation=velocity_representation)
+            if velocity_representation is not None
+            else {}
+        ),
```
flferretti, line 749: Is there any particular reason why you are unpacking a single item dictionary?
```suggestion
        velocity_representation=(
            velocity_representation if velocity_representation is not None else {}
        ),
```

src/jaxsim/api/model.py
flferretti, line 1: `step` does not use the argument `model`

flferretti (approved):
Thanks Diego, LGTM! I just wrote some minor comments

src/jaxsim/api/model.py
diegoferigo: Yep I'm aware and this is/was something wanted. There might be alternative `step*` methods in the future requiring `model` for internal computations. It was needed by the semi-implicit integrators, but probably with the most recent structure is no longer needed.

I leave a reminder in my code to check this sooner or later, likely when I'll upstream the semi-implicit integrators.

src/jaxsim/api/data.py
flferretti: Alright, thanks

traversaro (approved):

PR #108, opened 2024-03-12
Finalize the functional APIs and replace the OOP classes

diegoferigo:
This is a container PR merging the new structure of upcoming JaxSim release.

Extends the initial draft of the functional APIs introduced in #88 and #99 with the following PRs:

- #106 
- #101 
- #112
- #115 
- #116
- #117 
- #119
- #121 
- #120 
- #123

Note that tests running on CPU work fine, as in CI. For some reason, on my NVIDIA GeForce GTX 1650 Ti, testing the AD of a simulation step on a reduced ErgoCub model fails with the following error:

```
2024-03-28 15:40:01.469587: W external/xla/xla/service/gpu/runtime/support.cc:58] Intercepted XLA runtime error:
RESOURCE_EXHAUSTED: Failed to get module function:CUDA_ERROR_OUT_OF_MEMORY: out of memory
2024-03-28 15:40:01.469631: E external/xla/xla/pjrt/pjrt_stream_executor_client.cc:2732] Execution of replica 0 failed: INTERNAL: Failed to instantiate GPU graphs: CaptureGpuGraph failed (Failed to get module function:CUDA_ERROR_OUT_OF_MEMORY: out of memory; current tracing scope: fusion.173): INTERNAL: failed to capture gpu graph: run time error: custom call 'xla.gpu.func.launch' failed
```

The test suite on GPU can be executed as follows:

```bash
pytest -k "not test_ad_integration[ergocub_reduced]"
```

----
📚 Documentation preview 📚: https://jaxsim--108.org.readthedocs.build//108/

PR #114, opened 2024-03-14
Add `ruff`action and configuration

flferretti:
This PR adds the `ruff` action to CIs and pre-commit, together with the configuration that has been added to the `pyproject.toml`. Finally, a full lint using the current version has been performed.

----
📚 Documentation preview 📚: https://jaxsim--114.org.readthedocs.build//114/

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -14,12 +14,56 @@ version_file = "src/jaxsim/_version.py"
 line-length = 88
 
 [tool.isort]
-profile = "black"
 multi_line_output = 3
+profile = "black"
 
 [tool.pytest.ini_options]
+addopts = "-rsxX -v --strict-markers --forked"
 minversion = "6.0"
-addopts = "-rsxX -v --strict-markers"
 testpaths = [
-    "tests",
+  "tests",
 ]
+
+[tool.ruff]
+exclude = [
+  ".git",
+  ".pytest_cache",
+  ".ruff_cache",
+  ".vscode",
+  ".devcontainer",
+  "__pycache__",
+]
+preview = true
+
+target-version = "py311"
+
+[tool.ruff.lint]
+ignore = [
+  "B008", # Function call in default argument
+  "B024", # Abstract base class without abstract methods
+  "B904", # Raise without from inside exception
+  "B905", # Zip without explicit strict
+  "E402", # Module level import not at top of file
+  "E501", # Line too long
+  "E731", # Do not assign a `lambda` expression, use a `def`
+  "E741", # Ambiguous variable name
+  "F841", # Local variable is assigned to but never used
+  "I001", # Import block is unsorted or unformatted
+]
+select = [
+  "B",
+  "E",
+  "F",
+  "I",
+  "W",
+  "YTT",
+]
+
+[tool.ruff.lint.per-file-ignores]
+# Ignore `E402` (import violations) in all `__init__.py` files
+"**/{tests,docs,tools}/*" = ["E402"]
+"__init__.py" = ["F401"]
+# TODO: Remove these once deprecated code is removed
+"src/jaxsim/integrators/common.py" = ["E712"]
+"src/jaxsim/physics/model/physics_model.py" = ["B023"]
+"src/jaxsim/simulation/simulator_callbacks.py" = ["F821"]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -64,4 +64,5 @@ select = [
 [tool.ruff.lint.per-file-ignores]
 # Ignore `E402` (import violations) in all `__init__.py` files
 "**/{tests,docs,tools}/*" = ["E402"]
+"**/{tests}/*" = ["B007"]
 "__init__.py" = ["F401"]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,9 +1,9 @@
 [build-system]
 build-backend = "setuptools.build_meta"
 requires = [
-    "wheel",
-    "setuptools>=64",
-    "setuptools_scm[toml]>=8",
+  "wheel",
+  "setuptools>=64",
+  "setuptools_scm[toml]>=8",
 ]
 
 [tool.setuptools_scm]
@@ -18,11 +18,14 @@ multi_line_output = 3
 profile = "black"
 
 [tool.pytest.ini_options]
-addopts = "-rsxX -v --strict-markers --forked"
+addopts = "-rsxX -v --strict-markers"
 minversion = "6.0"
+preview = true
 testpaths = [
   "tests",
 ]
+
+target-version = "py311"
 
 [tool.ruff]
 exclude = [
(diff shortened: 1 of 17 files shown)


diegoferigo:
Is it ok for you @flferretti to wait for #108 to be ready? Right now, it is easier to rebase this PR on top of that one instead of the opposite.

pyproject.toml
```diff
@@ -14,12 +14,56 @@ version_file = "src/jaxsim/_version.py"
...
+  ".ruff_cache",
+  ".vscode",
+  ".devcontainer",
+  "__pycache__",
 ]
+preview = true
+
+target-version = "py311"
+
+[tool.ruff.lint]
```
diegoferigo, line 40:
```suggestion
# https://docs.astral.sh/ruff/rules/
[tool.ruff.lint]
```

pyproject.toml
```diff
@@ -14,12 +14,56 @@ version_file = "src/jaxsim/_version.py"
...
+]
+
+[tool.ruff.lint.per-file-ignores]
+# Ignore `E402` (import violations) in all `__init__.py` files
+"**/{tests,docs,tools}/*" = ["E402"]
+"__init__.py" = ["F401"]
+# TODO: Remove these once deprecated code is removed
+"src/jaxsim/integrators/common.py" = ["E712"]
+"src/jaxsim/physics/model/physics_model.py" = ["B023"]
+"src/jaxsim/simulation/simulator_callbacks.py" = ["F821"]
```
diegoferigo, line 69: These two files are no longer in `main`.

pyproject.toml
```diff
@@ -14,12 +14,56 @@ version_file = "src/jaxsim/_version.py"
...
+  "W",
+  "YTT",
+]
+
+[tool.ruff.lint.per-file-ignores]
+# Ignore `E402` (import violations) in all `__init__.py` files
+"**/{tests,docs,tools}/*" = ["E402"]
+"__init__.py" = ["F401"]
+# TODO: Remove these once deprecated code is removed
+"src/jaxsim/integrators/common.py" = ["E712"]
```
diegoferigo, line 67: Is this problem easy to fix?

src/jaxsim/api/data.py
```diff
@@ -417,7 +417,7 @@ def base_velocity(self) -> jtp.Vector:
     def generalized_position(self) -> tuple[jtp.Matrix, jtp.Vector]:
         """
         Get the generalized position
-        :math:`\mathbf{q} = ({}^W \mathbf{H}_B, \mathbf{s}) \in \text{SO}(3) \times \mathbb{R}^n`.
+        :math:`\\mathbf{q} = ({}^W \\mathbf{H}_B, \\mathbf{s}) \\in \text{SO}(3) \times \\mathbb{R}^n`.
```
diegoferigo, line 420: Can this be fixed using `r"""` as in other docstrings?

tests/conftest.py
```diff
@@ -176,14 +175,12 @@ def jaxsim_model_ergocub() -> js.model.JaxSimModel:
...
 
-    os_environ_original = os.environ.copy()
-
     try:
         os.environ["ROBOT_DESCRIPTION_COMMIT"] = "v0.7.1"
 
         import robot_descriptions.ergocub_description
     finally:
-        os.environ = os_environ_original
+        os.environ.clear()
```
diegoferigo, line 183: Doesn't `clear()` delete the entire dictionary? Here I want to restore the system one.

diegoferigo (changes requested):
Few questions and minor changes.

pyproject.toml
flferretti: Potentially yes, the error raised is:

```python
src/jaxsim/integrators/common.py:532:36: E712 Avoid equality comparisons to `True`; use `if rows_of_A_with_fsal:` for truth checks
    |
530 |         # possibly intermediate kᵢ derivative).
531 |         # Note that if multiple rows match (it should not), we return the first match.
532 |         return True, int(jnp.where(rows_of_A_with_fsal == True)[0].tolist()[0])
    |                                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^ E712
    |
    = help: Replace with `rows_of_A_with_fsal`
```

tests/conftest.py
flferretti: `clear()` deletes the environment variable defined during the process started by Python, we could eventually think to use `subprocess.Popen` to pass a custom env keys set to a subprocess

diegoferigo: It doesn't seem that it works as you described it:

```
❯ ipython
Python 3.11.8 | packaged by conda-forge | (main, Feb 16 2024, 20:53:32) [GCC 12.3.0]
Type 'copyright', 'credits' or 'license' for more information
IPython 8.22.2 -- An enhanced Interactive Python. Type '?' for help.

In [1]: import os

In [2]: os.environ["HOME"]
Out[2]: '/home/dferigo'

In [3]: os.environ["FOO"] = "bar"

In [4]: os.environ.clear()

In [5]: os.environ["HOME"]
---------------------------------------------------------------------------
KeyError                                  Traceback (most recent call last)
Cell In[5], line 1
----> 1 os.environ["HOME"]

File <frozen os>:679, in __getitem__(self, key)

KeyError: 'HOME'
```

What's the problem of the previous logic?

diegoferigo: Ow ok now I understand the problem. What about the following logic?

```python
    try:
        os.environ["ROBOT_DESCRIPTION_COMMIT"] = "v0.7.1"

        import robot_descriptions.ergocub_description
    finally:
        _ = os.environ.pop("ROBOT_DESCRIPTION_COMMIT", None)
```

flferretti: Sure, it makes sense

flferretti: Done in https://github.com/ami-iit/jaxsim/pull/114/commits/4cd9358b897b1efdd9621b0c32d31897920a0655

src/jaxsim/api/data.py
flferretti: Done in https://github.com/ami-iit/jaxsim/pull/114/commits/f7c63e4ff1a69bb6efd5d40b15544994c938e2ee

pyproject.toml
flferretti: Done in https://github.com/ami-iit/jaxsim/pull/114/commits/f7c63e4ff1a69bb6efd5d40b15544994c938e2ee

pyproject.toml
flferretti: Thanks, done in https://github.com/ami-iit/jaxsim/pull/114/commits/f7c63e4ff1a69bb6efd5d40b15544994c938e2ee

pyproject.toml
flferretti: Thanks, done in https://github.com/ami-iit/jaxsim/pull/114/commits/f7c63e4ff1a69bb6efd5d40b15544994c938e2ee

diegoferigo (approved):

PR #116, opened 2024-03-21
Improve utilities provided by `JaxsimDataclass`

diegoferigo:
When I first designed this class, I wanted to keep the `Mutability` enum private. However, in practice, there are multiple use-cases in which downstream projects need to use it.

This PR:

- Adjusts methods to expose publicly the `Mutability` enum.
- Reduces the number of copies, possibly improving the memory footprint of downstream code.
- Uses stricter checks to prevent JIT recompilations. Beyond pytree structure, now also the shape  and dtype of the leaves if checked, if they are arrays.
- Adds docstrings.

----
📚 Documentation preview 📚: https://jaxsim--116.org.readthedocs.build//116/


diegoferigo:
Moving this forward since it's not merging into `main`. @flferretti if you have comments, please leave them in #108.

PR #119, opened 2024-03-26
Fix `api.contact.in_contact`

flferretti:
The returned length of `links_in_contact` was always equal to the total number of links in the model, instead of being equal to the length of the input `link_names`

----
📚 Documentation preview 📚: https://jaxsim--119.org.readthedocs.build//119/

### DIFF
--- a/src/jaxsim/api/contact.py
+++ b/src/jaxsim/api/contact.py
@@ -7,6 +7,7 @@ import jaxsim.api as js
 import jaxsim.rbda
 import jaxsim.typing as jtp
 
+from . import link as Link
 from .common import VelRepr
 
 
@@ -177,7 +178,7 @@ def in_contact(
 
     link_names = link_names if link_names is not None else model.link_names()
 
-    if set(link_names) - set(model.link_names()) != set():
+    if (link_names := set(link_names)) and link_names.difference(model.link_names()):
         raise ValueError("One or more link names are not part of the model")
 
     W_p_Ci = collidable_point_positions(model=model, data=data)
@@ -190,11 +191,11 @@ def in_contact(
 
     links_in_contact = jax.vmap(
         lambda link_index: jnp.where(
-            model.kin_dyn_parameters.contact_parameters.body == link_index,
+            model.physics_model.gc.body == link_index,
             below_terrain,
             jnp.zeros_like(below_terrain, dtype=bool),
         ).any()
-    )(jnp.arange(model.number_of_links()))
+    )(Link.names_to_idxs(link_names=link_names, model=model))
 
     return links_in_contact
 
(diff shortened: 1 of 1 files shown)


diegoferigo (changes requested):
Thanks for having a look @flferretti. It is indeed wrongly returning always an array with $n_L$ elements. However, I fear your fix is not correct. I guess you need to create a boolean mask that selects the elements of `links_in_contact` corresponding to the passed `link_names`. Right now, you are selecting the first elements of `links_in_contact`, that would not correspond to `link_names`. Did I get it right?

src/jaxsim/api/contact.py
```diff
@@ -190,11 +191,11 @@ def in_contact(
 
     links_in_contact = jax.vmap(
         lambda link_index: jnp.where(
-            model.kin_dyn_parameters.contact_parameters.body == link_index,
+            model.physics_model.gc.body == link_index,
             below_terrain,
             jnp.zeros_like(below_terrain, dtype=bool),
         ).any()
-    )(jnp.arange(model.number_of_links()))
+    )(Link.names_to_idxs(link_names=link_names, model=model))
```
diegoferigo, line 198:
```suggestion
    )(js.link.names_to_idxs(link_names=link_names, model=model))
```

src/jaxsim/api/contact.py
```diff
@@ -7,6 +7,7 @@
 import jaxsim.rbda
 import jaxsim.typing as jtp
 
+from . import link as Link
```
diegoferigo, line 10:
```suggestion
```

src/jaxsim/api/contact.py
```diff
@@ -190,11 +191,11 @@ def in_contact(
 
     links_in_contact = jax.vmap(
         lambda link_index: jnp.where(
-            model.kin_dyn_parameters.contact_parameters.body == link_index,
+            model.physics_model.gc.body == link_index,
```
diegoferigo, line 194: The `functional` branch should no longer access `model.physics_model`.

```suggestion
            model.kin_dyn_parameters.contact_parameters.body == link_index,
```

diegoferigo (changes requested):

src/jaxsim/api/contact.py
diegoferigo: Tests are not failind because this function is not yet part of any test.

flferretti: Done in https://github.com/ami-iit/jaxsim/pull/119/commits/dd0d00b4c76315b48a2a096157ff6531e5e14037

src/jaxsim/api/contact.py
flferretti: Done in https://github.com/ami-iit/jaxsim/pull/119/commits/dd0d00b4c76315b48a2a096157ff6531e5e14037

src/jaxsim/api/contact.py
flferretti: Addressed in https://github.com/ami-iit/jaxsim/pull/119/commits/dd0d00b4c76315b48a2a096157ff6531e5e14037

PR #121, opened 2024-03-26
Optimize Jacobian algorithm

diegoferigo:
This PR enhances the computation of the Jacobians of all links. Before this PR, we computed in parallel with `jax.vmap` the free-floating left-trivialized Jacobian of all links ${}^L J_{W,L/B}$. Then, we were adjusting the input and output representations to match the desired ones.

This PR, instead of calling a parallelized version of `jaxsim.rbda.jacobian`, computes only once a _full_ doubly-left Jacobian ${}^B J_{W,\text{\\\_}/B}$ (note that there is a $\text{\\\_}$ instead of $L$), defined as the free-floating Jacobian with all rows filled. The free-floating doubly-left Jacobian of the i-th link ${}^B J_{W,L/B}$ is then computed by filtering the columns of the full Jacobian with the support parent array $\kappa(i)$ of the link, that sets to zero the columns corresponding to all links not part of the path $\pi_B(L)$.

All of this allows to compute the Jacobians of all links by only vmapping the full Jacobian using $\kappa(i)$, therefore the expensive algorithm to compute the full Jacobian is executed only once. The vmapped operation is just a column filtering, therefore its cost is almost zero.

This enhancement should speed up the computation of `jaxsim.api.model.forward_dynamics_crb`. While at this stage is always better to use `jaxsim.api.model.forward_dynamics_aba`, the CRB version might become useful in all experiments that extend the forward dynamics with e.g. actuators, friction models, or any other stateless second-order dynamics elements[^1]. In these cases, operating on ABA could become quite difficult, and prototyping using the CRB version might be pretty helpful.

cc @traversaro @DanielePucci 

[^1]: Under the assumption that they do not extend the integrated state vector. In that case, things become more complex since also the integrator has to properly updated.

----
📚 Documentation preview 📚: https://jaxsim--121.org.readthedocs.build//121/

### DIFF
--- a/src/jaxsim/api/link.py
+++ b/src/jaxsim/api/link.py
@@ -233,62 +233,61 @@ def jacobian(
         output_vel_repr if output_vel_repr is not None else data.velocity_representation
     )
 
-    # Compute the doubly left-trivialized free-floating jacobian.
-    L_J_WL_B = jaxsim.rbda.jacobian(
+    # Compute the doubly-left free-floating full jacobian.
+    B_J_full_WX_B, B_H_Li = jaxsim.rbda.jacobian_full_doubly_left(
         model=model,
-        link_index=link_index,
         joint_positions=data.joint_positions(),
     )
 
-    # We want to return the Jacobian O_J_WL_I, where O is the representation of the
-    # output 6D velocity, and I is the representation of the input generalized velocity.
+    # Compute the actual doubly-left free-floating jacobian of the link.
+    κ = model.kin_dyn_parameters.support_body_array_bool[link_index]
+    B_J_WL_B = jnp.hstack([jnp.ones(5), κ]) * B_J_full_WX_B
 
-    # Change the input representation matching the one of data.
+    # Adjust the input representation such that `J_WL_I @ I_ν`.
     match data.velocity_representation:
+        case VelRepr.Inertial:
+            W_H_B = data.base_transform()
+            B_X_W = jaxlie.SE3.from_matrix(W_H_B).inverse().adjoint()
+            B_J_WL_I = B_J_WL_W = B_J_WL_B @ jax.scipy.linalg.block_diag(
+                B_X_W, jnp.eye(model.dofs())
+            )
 
         case VelRepr.Body:
-            L_J_WL_I = L_J_WL_B
-
-        case VelRepr.Inertial:
-
-            W_H_B = data.base_transform()
-
-            B_X_W = jaxlie.SE3.from_matrix(W_H_B).inverse().adjoint()
-            B_T_W = jax.scipy.linalg.block_diag(B_X_W, jnp.eye(model.dofs()))
-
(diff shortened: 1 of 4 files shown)


traversaro (approved):
Cool! Did you did any benchmark on the difference before and after this change?

diegoferigo:
I was running them when you were typing your comment :)

Here below a benchmark on a 58-DoFs ErgoCub model on CPU.

## Before

```
In [2]: %timeit -r1 -n1 _ = js.model.generalized_free_floating_jacobian(model, data0)
28 s ± 0 ns per loop (mean ± std. dev. of 1 run, 1 loop each)

In [3]: %timeit -r 10 -n 1000 _ = js.model.generalized_free_floating_jacobian(model, data0)
1.37 ms ± 82.6 µs per loop (mean ± std. dev. of 10 runs, 1,000 loops each)

In [4]: %timeit -r1 -n1 _ = js.model.forward_dynamics_crb(model, data0)
1min 27s ± 0 ns per loop (mean ± std. dev. of 1 run, 1 loop each)

In [5]: %timeit -r10 -n1000 _ = js.model.forward_dynamics_crb(model, data0)
1.99 ms ± 69.1 µs per loop (mean ± std. dev. of 10 runs, 1,000 loops each)
```

## After

```
In [2]: %timeit -r1 -n1 _ = js.model.generalized_free_floating_jacobian(model, data0)
9.5 s ± 0 ns per loop (mean ± std. dev. of 1 run, 1 loop each)

In [3]: %timeit -r 10 -n 1000 _ = js.model.generalized_free_floating_jacobian(model, data0)
185 µs ± 16.4 µs per loop (mean ± std. dev. of 10 runs, 1,000 loops each)

In [4]: %timeit -r1 -n1 _ = js.model.forward_dynamics_crb(model, data0)
55.6 s ± 0 ns per loop (mean ± std. dev. of 1 run, 1 loop each)

In [5]: %timeit -r10 -n1000 _ = js.model.forward_dynamics_crb(model, data0)
896 µs ± 51.5 µs per loop (mean ± std. dev. of 10 runs, 1,000 loops each)
```

## Overview

- Compiling `generalized_free_floating_jacobian` is now 3x faster.
- Compiling `forward_dynamics_crb` is now 1.5x faster.
- Running `generalized_free_floating_jacobian` is now 7x faster.
- Running `forward_dynamics_crb` is now 2x faster.

diegoferigo:
In other words, on such a large model, with this change we can compute FD with CRB in less than 1ms with JAX running on CPU. Not too bad. As a comparison, the equivalent computation with ABA runs 3x faster:

```
In [15]: %timeit _ = js.model.forward_dynamics_aba(model, data0)
346 µs ± 6.74 µs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)
```

diegoferigo:
cc @ami-iit/vertical_control-oriented-learning

PR #135, opened 2024-04-04
Fix `RodModelToMjcf.convert` when no `cameras` argument is passed

diegoferigo:
Regression introduced in #133.

----
📚 Documentation preview 📚: https://jaxsim--135.org.readthedocs.build//135/

### DIFF
--- a/src/jaxsim/mujoco/loaders.py
+++ b/src/jaxsim/mujoco/loaders.py
@@ -476,7 +476,7 @@ class RodModelToMjcf:
         )
 
         # Add user-defined camera
-        cameras = cameras if cameras is not None else {}
+        cameras = cameras if cameras is not None else []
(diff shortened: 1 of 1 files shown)


flferretti (approved):
Thanks! I wasn't expecting that issue as in theory `cameras` should be converted to a list of an empty dictionary. We should probably add some tests on the visualizer for the next release

diegoferigo:
> Thanks! I wasn't expecting that issue as in theory `cameras` should be converted to a list of an empty dictionary. We should probably add some tests on the visualizer for the next release

The problem is that a list of empty dictionaries like `[{}]` enters the `for` loop and triggers the construction of a `MujocoCamera` instance. This PR ensures that no `MujocoCamera` is created when `cameras=None`.

Issue #136, opened 2024-04-08
Performing a simulation enforcing JAX to run in 32-bit precision overflows after 4.29 seconds


diegoferigo:
JaxSim, by default, [enforces](https://github.com/ami-iit/jaxsim/blob/10c719984c0b600afd165d65a25f215d8bc28a75/src/jaxsim/__init__.py#L11-L20) JAX to run in [64-bit precision](https://jax.readthedocs.io/en/latest/notebooks/Common_Gotchas_in_JAX.html#double-64bit-precision). This is necessary to execute our RBDAs and compare the results to third-party C++ libraries.

As explained in #109, executing JaxSim with 32-bit precision might result to a significant speedup, especially on hardware accelerators like GPUs. Furthermore, TPUs only support 32-bits (https://github.com/google/jax/issues/9862), therefore we need to ensure that JaxSim runs with this precision if we want to exploit this hardware.

**JaxSim is _almost_ 32-bit compatible, the only caveats originate on how we handle the simulation time:**

https://github.com/ami-iit/jaxsim/blob/10c719984c0b600afd165d65a25f215d8bc28a75/src/jaxsim/api/data.py#L42-L44

When running in 32-bit precision, `time_ns` is downcasted as `jnp.uint32`. Unfortunately, the range of this data type is problematic when it describes nanoseconds. In fact:

```
In [1]: jnp.iinfo(jnp.uint32)
Out[1]: iinfo(min=0, max=4294967295, dtype=uint32)
```

The maximum time we can simulate without overflowing is approximately 4.29 seconds. When running a simulation longer than that, the simulator hangs, requiring to press <kbd>Ctrl</kbd>+<kbd>C</kbd>.

Something else to notice is that a time is seconds represented as a 32-bit float will have the following limits:

```
In [2]: jnp.finfo(jnp.float32)
Out[2]: finfo(resolution=1e-06, min=-3.4028235e+38, max=3.4028235e+38, dtype=float32)
```

Therefore, in this case we cannot achieve a perfect granularity up to 1 nanosecond (it's pointless going under that for rigid body simulations, already 10-100 ns is small enough).

---

We should investigate alternative representations of the simulation time.

As a consequence, all functions accepting the current time should be ported to this new representation, that will be something more complex than a plain `jax.Array`:

[...]

PR #152, opened 2024-05-15
Mark the use of Colab for the examples as broken

flferretti:
From https://github.com/ami-iit/rod/issues/32

----
📚 Documentation preview 📚: https://jaxsim--152.org.readthedocs.build//152/


traversaro (approved):
I would prefer to just get rid of the colab links as listing as Option 1 something broken is confusing, and colab links are present before the `Option 1` section, but for sure it is better then before!

diegoferigo:
If it is not working, I'd remove the whole Colab option. We can add it back as soon as it gets fixed.

diegoferigo (approved):

diegoferigo:
Regardless, I believe people might be interested in using Colab for trying running JaxSim on TPUs beyond GPUs. Something to keep in mind is that TPU support is blocked by #136. I have a fix in WIP, but then got caught up on other things with higher priority.

PR #154, opened 2024-05-20
Restore Colab support for example notebooks

flferretti:
This PR restore the support for Google Colab for the example notebooks previously removed in #152. 

----
📚 Documentation preview 📚: https://jaxsim--154.org.readthedocs.build//154/


flferretti:
I had a small issue on the visualization that has been solved in https://github.com/ami-iit/jaxsim/pull/154/commits/bea65bc1a7ac331ae52a2219fe02e58fe3406a30. Apparently in Colab, `libGL` requires a `DISPLAY` environment variable to be set. For this reason I used `xvfb` to create a dummy one.

Ready for review

traversaro:
> Apparently in Colab, `libGL` requires a `DISPLAY` environment variable to be set. For this reason I used `xvfb` to create a dummy one.

That is strange, as long as you use `egl` that should not be necessary. Not blocking the PR, but what was the error?

flferretti:
> > Apparently in Colab, `libGL` requires a `DISPLAY` environment variable to be set. For this reason I used `xvfb` to create a dummy one.
> 
> That is strange, as long as you use `egl` that should not be necessary. Not blocking the PR, but what was the error?

image

traversaro:
Thanks, I wonder why this is not happening in official mujoco colabs. Anyhow, no problem.

xela-95 (approved):

flferretti:
> Thanks, I wonder why this is not happening in official mujoco colabs. Anyhow, no problem.

I have no idea to be honest, I have also seen that in Mujoco it works

traversaro:
> > Thanks, I wonder why this is not happening in official mujoco colabs. Anyhow, no problem.
> 
> I have no idea to be honest, I have also seen that in Mujoco it works

Probably the difference is due to the fact that `glfw` is also installed or somehow is selected as backend?

PR #156, opened 2024-05-20
Add mesh support

lorycontixd:
This pull requests aims to extend Jaxsim's parsing functionalities on robot descriptions to meshes.
It does so by using third-party library [trimesh](https://github.com/mikedh/trimesh) to parse the mesh file, and custom algorithm `create_mesh_collisions` to wrap a mesh with collidable points.

----
📚 Documentation preview 📚: https://jaxsim--156.org.readthedocs.build//156/

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,7 +1,0 @@
-{
-    "python.testing.pytestArgs": [
-        "tests"
-    ],
-    "python.testing.unittestEnabled": false,
-    "python.testing.pytestEnabled": true
-}
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,7 @@
+{
+    "python.testing.pytestArgs": [
+        "tests"
+    ],
+    "python.testing.unittestEnabled": false,
+    "python.testing.pytestEnabled": true
+}
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -56,7 +56,6 @@ dependencies = [
     "rod >= 0.3.3",
     "typing_extensions ; python_version < '3.12'",
     "trimesh",
-    "manifold3d",
 ]
 
 [project.optional-dependencies]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -69,9 +69,7 @@ testing = [
     "idyntree >= 12.2.1",
     "pytest >=6.0",
     "pytest-icdiff",
-    "robot-descriptions",
-    "networkx",
-    "lark"
+    "robot-descriptions"
 ]
 viz = [
     "lxml",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -56,6 +56,7 @@ dependencies = [
     "rod >= 0.3.3",
     "typing_extensions ; python_version < '3.12'",
     "trimesh",
+    "manifold3d",
 ]
 
 [project.optional-dependencies]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -55,6 +55,7 @@ dependencies = [
     "qpax",
     "rod >= 0.3.3",
     "typing_extensions ; python_version < '3.12'",
+    "trimesh",
 ]
 
 [project.optional-dependencies]
@@ -68,6 +69,8 @@ testing = [
     "pytest >=6.0",
     "pytest-icdiff",
     "robot-descriptions",
+    "networkx",
+    "lark"
 ]
 viz = [
     "lxml",
--- a/src/jaxsim/parsers/descriptions/collision.py
+++ b/src/jaxsim/parsers/descriptions/collision.py
@@ -154,3 +154,16 @@ class SphereCollision(CollisionShape):
             return False
 
         return hash(self) == hash(other)
+
+
+@dataclasses.dataclass
+class MeshCollision(CollisionShape):
+
+    center: npt.NDArray[np.float64]
+
+    def __eq__(self, other: Any) -> bool:
+        if not isinstance(other, MeshCollision):
+            return False
+        return len(self.collidable_points) == len(
+            other.collidable_points
+        ) and super().__eq__(other)
--- a/src/jaxsim/parsers/descriptions/collision.py
+++ b/src/jaxsim/parsers/descriptions/collision.py
@@ -158,12 +158,11 @@ class SphereCollision(CollisionShape):
 
 @dataclasses.dataclass
 class MeshCollision(CollisionShape):
-
-    center: npt.NDArray[np.float64]
+    center: npt.NDArray
 
     def __eq__(self, other: Any) -> bool:
         if not isinstance(other, MeshCollision):
             return False
         return len(self.collidable_points) == len(
             other.collidable_points
-        ) and super().__eq__(other)
+        ) and super().__eq__(other) and (self.center == other.center).all()
--- a/src/jaxsim/parsers/descriptions/collision.py
+++ b/src/jaxsim/parsers/descriptions/collision.py
@@ -163,6 +163,8 @@ class MeshCollision(CollisionShape):
     def __eq__(self, other: Any) -> bool:
         if not isinstance(other, MeshCollision):
             return False
-        return len(self.collidable_points) == len(
-            other.collidable_points
-        ) and super().__eq__(other) and (self.center == other.center).all()
+        return (
+            len(self.collidable_points) == len(other.collidable_points)
+            and super().__eq__(other)
+            and (self.center == other.center).all()
+        )
--- a/src/jaxsim/parsers/descriptions/collision.py
+++ b/src/jaxsim/parsers/descriptions/collision.py
@@ -158,13 +158,19 @@ class SphereCollision(CollisionShape):
 
 @dataclasses.dataclass
 class MeshCollision(CollisionShape):
-    center: npt.NDArray
+    center: jtp.VectorLike
 
-    def __eq__(self, other: Any) -> bool:
+    def __hash__(self) -> int:
+        return hash(
+            (
+                hash(super()),
+                hash(tuple(self.center.tolist())),
+                hash(self.collidable_points),
+            )
+        )
+
+    def __eq__(self, other: MeshCollision) -> bool:
         if not isinstance(other, MeshCollision):
             return False
-        return (
-            len(self.collidable_points) == len(other.collidable_points)
-            and super().__eq__(other)
-            and (self.center == other.center).all()
-        )
+
+        return hash(self) == hash(other)
--- a/src/jaxsim/parsers/descriptions/collision.py
+++ b/src/jaxsim/parsers/descriptions/collision.py
@@ -163,7 +163,6 @@ class MeshCollision(CollisionShape):
     def __hash__(self) -> int:
         return hash(
             (
-                hash(super()),
                 hash(tuple(self.center.tolist())),
                 hash(self.collidable_points),
             )
--- a/src/jaxsim/parsers/rod/meshes.py
+++ b/src/jaxsim/parsers/rod/meshes.py
@@ -1,9 +1,11 @@
-import trimesh
+from abc import ABC, abstractmethod
+from typing import List, Sequence
+
 import numpy as np
 import rod
-from typing import Sequence, List
+import trimesh
+
 from jaxsim import logging
-from abc import ABC, abstractmethod
 
 
 def parse_object_mapping_object(obj) -> trimesh.Trimesh:
--- a/src/jaxsim/parsers/rod/meshes.py
+++ b/src/jaxsim/parsers/rod/meshes.py
@@ -1,0 +1,80 @@
+import trimesh
+import numpy as np
+
+
+def extract_points_vertex_extraction(mesh: trimesh.Trimesh) -> np.ndarray:
+    """Extracts the points of a mesh using the vertices of the mesh as colliders.
+
+    Args:
+        mesh: The mesh to extract the points from.
+
+    Returns:
+        The points of the mesh.
+    """
+    return mesh.vertices
+
+
+def extract_points_random_surface_sampling(
+    mesh: trimesh.Trimesh, num_points: int
+) -> np.ndarray:
+    """Extracts the points of a mesh by sampling the surface of the mesh randomly.
+
+    Args:
+        mesh: The mesh to extract the points from.
+        num_points: The number of points to sample.
+
+    Returns:
+        The points of the mesh.
+    """
+    return mesh.sample(num_points)
+
+
+def extract_points_uniform_surface_sampling(
+    mesh: trimesh.Trimesh, num_points: int
+) -> np.ndarray:
+    """Extracts the points of a mesh by sampling the surface of the mesh uniformly.
+
+    Args:
+        mesh: The mesh to extract the points from.
+        num_points: The number of points to sample.
+
+    Returns:
+        The points of the mesh.
+    """
+    return trimesh.sample.sample_surface_even(mesh=mesh, count=num_points)
+
+
+def extract_points_aap(
+    mesh: trimesh.Trimesh,
+    aap_axis: str,
+    aap_value: float,
+    aap_direction: str,
+) -> np.ndarray:
+    """Extracts the points of a mesh that are on one side of an axis-aligned plane (AAP).
+
+    Args:
+        mesh: The mesh to extract the points from.
+        aap_axis: The axis of the AAP.
+        aap_value: The value of the AAP.
+        aap_direction: The direction of the AAP.
+
+    Returns:
+        The points of the mesh that are on one side of the AAP.
+    """
+    if aap_direction == "higher":
+        aap_operator = np.greater
+    elif aap_direction == "lower":
+        aap_operator = np.less
+    else:
+        raise ValueError("Invalid direction for axis-aligned plane")
+
+    if aap_axis == "x":
+        points = mesh.vertices[aap_operator(mesh.vertices[:, 0], aap_value)]
+    elif aap_axis == "y":
+        points = mesh.vertices[aap_operator(mesh.vertices[:, 1], aap_value)]
+    elif aap_axis == "z":
+        points = mesh.vertices[aap_operator(mesh.vertices[:, 2], aap_value)]
+    else:
+        raise ValueError("Invalid axis for axis-aligned plane")
+
+    return points
--- a/src/jaxsim/parsers/rod/meshes.py
+++ b/src/jaxsim/parsers/rod/meshes.py
@@ -1,80 +1,202 @@
 import trimesh
 import numpy as np
+import rod
 
 
-def extract_points_vertex_extraction(mesh: trimesh.Trimesh) -> np.ndarray:
-    """Extracts the points of a mesh using the vertices of the mesh as colliders.
-
-    Args:
-        mesh: The mesh to extract the points from.
-
-    Returns:
-        The points of the mesh.
-    """
-    return mesh.vertices
-
-
-def extract_points_random_surface_sampling(
-    mesh: trimesh.Trimesh, num_points: int
-) -> np.ndarray:
-    """Extracts the points of a mesh by sampling the surface of the mesh randomly.
-
-    Args:
-        mesh: The mesh to extract the points from.
-        num_points: The number of points to sample.
-
-    Returns:
-        The points of the mesh.
-    """
-    return mesh.sample(num_points)
-
-
-def extract_points_uniform_surface_sampling(
-    mesh: trimesh.Trimesh, num_points: int
-) -> np.ndarray:
-    """Extracts the points of a mesh by sampling the surface of the mesh uniformly.
-
-    Args:
-        mesh: The mesh to extract the points from.
-        num_points: The number of points to sample.
-
-    Returns:
-        The points of the mesh.
-    """
-    return trimesh.sample.sample_surface_even(mesh=mesh, count=num_points)
-
-
-def extract_points_aap(
-    mesh: trimesh.Trimesh,
-    aap_axis: str,
-    aap_value: float,
-    aap_direction: str,
-) -> np.ndarray:
-    """Extracts the points of a mesh that are on one side of an axis-aligned plane (AAP).
-
-    Args:
-        mesh: The mesh to extract the points from.
-        aap_axis: The axis of the AAP.
-        aap_value: The value of the AAP.
-        aap_direction: The direction of the AAP.
-
-    Returns:
-        The points of the mesh that are on one side of the AAP.
-    """
-    if aap_direction == "higher":
-        aap_operator = np.greater
-    elif aap_direction == "lower":
-        aap_operator = np.less
+def parse_object_mapping_object(obj) -> trimesh.Trimesh:
+    if isinstance(obj, trimesh.Trimesh):
+        return obj
+    elif isinstance(obj, dict):
+        if "type" not in obj:
+            raise ValueError("Object type not specified")
+        if obj["type"] == "box":
+            if "extents" not in obj:
+                raise ValueError("Box extents not specified")
+            return trimesh.creation.box(extents=obj["extents"])
+        elif obj["type"] == "sphere":
+            if "radius" not in obj:
+                raise ValueError("Sphere radius not specified")
+            return trimesh.creation.icosphere(subdivisions=4, radius=obj["radius"])
+        else:
+            raise ValueError(f"Invalid object type {obj['type']}")
+    elif isinstance(obj, rod.builder.primitive_builder.PrimitiveBuilder):
+        raise NotImplementedError("PrimitiveBuilder not implemented")
     else:
-        raise ValueError("Invalid direction for axis-aligned plane")
+        raise ValueError("Invalid object type")
 
-    if aap_axis == "x":
-        points = mesh.vertices[aap_operator(mesh.vertices[:, 0], aap_value)]
-    elif aap_axis == "y":
-        points = mesh.vertices[aap_operator(mesh.vertices[:, 1], aap_value)]
-    elif aap_axis == "z":
-        points = mesh.vertices[aap_operator(mesh.vertices[:, 2], aap_value)]
-    else:
-        raise ValueError("Invalid axis for axis-aligned plane")
 
-    return points
+class MeshMapping:
+    @staticmethod
+    def vertex_extraction(mesh: trimesh.Trimesh) -> np.ndarray:
+        """Extracts the points of a mesh using the vertices of the mesh as colliders.
+
+        Args:
+            mesh: The mesh to extract the points from.
+
+        Returns:
+            The points of the mesh.
+        """
+        return mesh.vertices
+
+    @staticmethod
+    def random_surface_sampling(mesh: trimesh.Trimesh, num_points: int) -> np.ndarray:
+        """Extracts the points of a mesh by sampling the surface of the mesh randomly.
+
+        Args:
+            mesh: The mesh to extract the points from.
+            num_points: The number of points to sample.
+
+        Returns:
+            The points of the mesh.
+        """
+        return mesh.sample(num_points)
+
+    @staticmethod
+    def uniform_surface_sampling(mesh: trimesh.Trimesh, num_points: int) -> np.ndarray:
+        """Extracts the points of a mesh by sampling the surface of the mesh uniformly.
+
+        Args:
+            mesh: The mesh to extract the points from.
+            num_points: The number of points to sample.
+
+        Returns:
+            The points of the mesh.
+        """
+        return trimesh.sample.sample_surface_even(mesh=mesh, count=num_points)
+
+    @staticmethod
+    def aap(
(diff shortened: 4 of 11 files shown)


flferretti:
@lorycontixd do you prefer a squash and merge or can you please clean the commit history?

.vscode/settings.json
flferretti, line 1: This should be removed

src/jaxsim/parsers/rod/utils.py
```diff
@@ -1,15 +1,23 @@
...
 
 import jaxsim.typing as jtp
 from jaxsim.math.inertia import Inertia
 from jaxsim.parsers import descriptions
 
 
+class MeshMappingMethods(enum.Enum):
+    VertexExtraction = (0,)
+    RandomSurfaceSampling = (1,)
+    UniformSurfaceSampling = 2
```
flferretti, line 18: You can leverage `enum.IntEnum` here (see https://docs.python.org/3/library/enum.html#enum.IntEnum)
```suggestion
@enum.unique
class MeshMappingMethods(enum.IntEnum):
    VertexExtraction = enum.auto()
    RandomSurfaceSampling = enum.auto()
    UniformSurfaceSampling = enum.auto()
```

.vscode/settings.json
```diff
@@ -0,0 +1,7 @@
+{
+    "python.testing.pytestArgs": [
+        "tests"
+    ],
+    "python.testing.unittestEnabled": false,
+    "python.testing.pytestEnabled": true
+}
```
diegoferigo, line 7: please keep your personal IDE configuration out from the committed files

diegoferigo, line 7: please keep your personal IDE configuration out from the committed files

setup.cfg
```diff
@@ -61,6 +61,7 @@ install_requires =
     jax_dataclasses >= 1.4.0
     pptree
     rod >= 0.2.0
+    trimesh
```
diegoferigo, line 64: Can you please add this new dependency in the environment.yml file for conda-based installations?

src/jaxsim/parsers/rod/parser.py
```diff
@@ -305,6 +307,39 @@ def extract_model_data(
...
+                logging.debug(
+                    f"===> Loading mesh {collision.geometry.mesh.uri} with scale {collision.geometry.mesh.scale}, file type '{_file_type}'"
+                )
+                mesh = trimesh.load(
+                    file_obj=open(file_obj, "rb"),
+                    file_type=_file_type,
+                )
+
+                if isinstance(mesh, trimesh.Trimesh):
+                    mesh.apply_scale(collision.geometry.mesh.scale)
```
diegoferigo, line 329: Can you move this logic inside `utils.create_mesh_collision`?

tests/test_parser_mesh.py
flferretti, line 1: Empty. Would you like to remove this?

flferretti (changes requested):
Thanks @lorycontixd! I wrote some initial comments

diegoferigo (changes requested):
Thanks for your contribution @lorycontixd, it's a great start. First minor changes, I'll review the logic in more details after these ones are addressed.

src/jaxsim/parsers/rod/parser.py
```diff
@@ -326,6 +326,18 @@ def extract_model_data(
...
                 collisions.append(sphere_collision)
 
+            if (
+                collision.geometry.mesh is not None
+                and collision.geometry.mesh.uri is not None
+            ):
+                mesh_collision = utils.create_mesh_collision(
+                    collision=collision,
+                    link_description=links_dict[link.name],
+                    mesh_uri=collision.geometry.mesh.uri,
```
diegoferigo, line 336: This is already part of the `collision` argument, can you extract it and use it directly inside `create_mesh_collision`? I would also remove the check if the uri is None inside this function, and raise an exception if it is. I don't think we want to support any mesh without a defined uri, and failing silently can lead to user surprise.

src/jaxsim/parsers/rod/utils.py
```diff
@@ -206,3 +217,62 @@ def fibonacci_sphere(samples: int) -> npt.NDArray:
...
+    link_description: descriptions.LinkDescription,
+    mesh_uri: str,
+    method: MeshMappingMethods = MeshMappingMethods.VertexExtraction,
+    nsamples: int = 1000,
+) -> descriptions.MeshCollision:
+    file_obj = resolve_local_uri(uri=collision.geometry.mesh.uri)
+    try:
+        _file_type = str(file_obj).split(".")[-1]
+    except ValueError as e:
+        raise e(f"Failed to get file type from {file_obj}")
```
diegoferigo, line 233: If I recall, trimesh already fails if it receives an extension that is not supported. Can you just pass to trimesh `file_type=pathlib.Path(self.mesh_path).suffix.replace(".", "")`?

src/jaxsim/parsers/rod/utils.py
```diff
@@ -206,3 +217,62 @@ def fibonacci_sphere(samples: int) -> npt.NDArray:
...
+    logging.debug(
+        f"===> Loading mesh {collision.geometry.mesh.uri} with scale {collision.geometry.mesh.scale}, file type '{_file_type}'"
+    )
+    mesh = trimesh.load(
+        file_obj=open(file_obj, "rb"),
+        file_type=_file_type,
+    )
```
diegoferigo, line 240: Is the file correctly close afterwards? I'd suggest to use a context manager guarding these lines.

src/jaxsim/parsers/rod/utils.py
```diff
@@ -206,3 +217,62 @@ def fibonacci_sphere(samples: int) -> npt.NDArray:
...
+    # Checking if mesh is empty
+    if isinstance(mesh, trimesh.Trimesh):
+        mesh.apply_scale(collision.geometry.mesh.scale)
+    else:
+        logging.info(f"Mesh '{collision.geometry.mesh.uri}' is empty/.")
+        return
```
diegoferigo, line 246: What are the circumstances in which trimesh exits without raising any exception and returns an empty `mesh` object?

diegoferigo (changes requested):
Comments on trimesh usage. Almost good on my side, I'll do a final check tomorrow.

setup.cfg
diegoferigo: This has not yet been resoved

diegoferigo:
For the records, this is the function called to load a mesh in trimesh:

https://trimesh.org/trimesh.html#trimesh.load

src/jaxsim/parsers/rod/utils.py
diegoferigo: I see from https://trimesh.org/trimesh.html#trimesh.load that `trimesh.load` accepts a `pathlib.Path`. Can you use it instead of a file object? Furthermore, since the URI can only point to mesh files, what about using `trimesh.load_mesh`?

diegoferigo: I suspect that using `pathlib.Path` objects as suggested, trimesh might be smart enough to automatically detect the extension without the need to explicitly extract the suffix.

flferretti:
@diegoferigo before merging this, do we want to set meshes support as default or should it be activated using an environment variable?

diegoferigo:
> @diegoferigo before merging this, do we want to set meshes support as default or should it be activated using an environment variable?

Yes I need to properly assess the implications of merging this PR on existing appplications. Automatically adding the collidable points of all meshes is a huge change of behavior and it will make existing simulations unbearably slow.

lorycontixd:
@flferretti @flferretti As of now, random point sampling from a mesh is supported. I think it would be a good idea if I tried to benchmark different values of sampling points and find a good tradeoff between performance and simulation results.

setup.cfg
```diff
@@ -74,6 +75,7 @@ style =
     pre-commit
 testing =
     idyntree >= 12.2.1
+    networkx
```
diegoferigo, line 78: What is this new dependency used for?

src/jaxsim/parsers/descriptions/collision.py
```diff
@@ -119,3 +119,16 @@ class SphereCollision(CollisionShape):
 
     def __eq__(self, other):
         return (self.center == other.center).all() and super().__eq__(other)
+
+
+@dataclasses.dataclass
+class MeshCollision(CollisionShape):
+
+    center: npt.NDArray[np.float64]
```
diegoferigo, line 127: Let's keep this simple, we never use explicit types throughout our codebase.

```suggestion
    center: npt.NDArray
```

src/jaxsim/parsers/descriptions/collision.py
```diff
@@ -119,3 +119,16 @@ class SphereCollision(CollisionShape):
...
+
+    def __eq__(self, other: Any) -> bool:
+        if not isinstance(other, MeshCollision):
+            return False
+        return len(self.collidable_points) == len(
+            other.collidable_points
+        ) and super().__eq__(other)
```
diegoferigo, line 134: Here the comparison of the `MeshCollision.center` attribute seems missing.

src/jaxsim/parsers/rod/parser.py
```diff
@@ -326,6 +326,15 @@ def extract_model_data(
...
                 collisions.append(sphere_collision)
 
+            if collision.geometry.mesh is not None:
+                mesh_collision = utils.create_mesh_collision(
+                    collision=collision,
+                    link_description=links_dict[link.name],
+                    method=utils.MeshMappingMethods.VertexExtraction,
+                )
+                if mesh_collision is not None:
+                    collisions.append(mesh_collision)
```
diegoferigo, line 336: Can this ever happen? I guess that in the case `utils.create_mesh_collision` fails, an exception is raised.

src/jaxsim/parsers/rod/utils.py
```diff
@@ -206,3 +218,54 @@ def fibonacci_sphere(samples: int) -> npt.NDArray:
...
+    method: MeshMappingMethods = MeshMappingMethods.VertexExtraction,
+    nsamples: int = 1000,
+) -> descriptions.MeshCollision:
+    file = pathlib.Path(resolve_local_uri(uri=collision.geometry.mesh.uri))
+    _file_type = file.suffix.replace(".", "")
+    mesh = trimesh.load_mesh(file, file_type=_file_type)
+
+    if mesh.is_empty:
+        logging.warning(f"Mesh {collision.geometry.mesh.uri} is empty, ignoring it")
+        return
```
diegoferigo, line 235: When can this happpen? The `mesh` variable is populated by `trimesh`. If the parsing fails, I'd expect that trimesh raises. If the library does not raise, I'd change the code as follows:

```suggestion
    if mesh.is_empty:
        raise RuntimeError(f"Failed to process '{file}' with trimesh")
```

src/jaxsim/parsers/rod/utils.py
```diff
@@ -206,3 +218,54 @@ def fibonacci_sphere(samples: int) -> npt.NDArray:
...
+            points = mesh.sample(nsamples)
+        case MeshMappingMethods.UniformSurfaceSampling:
+            raise NotImplementedError("Uniform surface sampling not implemented")
+        case _:
+            raise ValueError("Invalid mesh mapping method")
+
+    points = mesh.vertices
+    H = (
+        collision.pose.transform() if collision.pose is not None else np.eye(4)
+    )  # pose of the collision object
```
diegoferigo, line 254: nit: can you move comments in a new line before the variable assignment instead of in the end of its line?

diegoferigo (changes requested):
Almost ready, last comments on the changes before started assessing the new default behavior of mesh collisions in JaxSim.

src/jaxsim/parsers/rod/utils.py
flferretti: This happens when the ASCII-STL file is empty, e.g. for the root link mesh of ErgoCub when downloaded using `robot_descriptions` (probably a bug upstream)

src/jaxsim/parsers/rod/utils.py
```diff
@@ -206,3 +218,54 @@ def fibonacci_sphere(samples: int) -> npt.NDArray:
...
+    center_of_collision_wrt_link = (H @ np.hstack([0, 0, 0, 1.0]))[
+        0:-1
+    ]  # @ = matrix multiplication, hstack = stack arrays in sequence horizontally => center of the collision object
```
flferretti, line 257: Leftover?
```suggestion
    ]
```

tests/conftest.py
```diff
@@ -182,12 +182,22 @@ def jaxsim_model_ergocub() -> js.model.JaxSimModel:
     finally:
         _ = os.environ.pop("ROBOT_DESCRIPTION_COMMIT", None)
 
+    from robot_descriptions._package_dirs import get_package_dirs
+
+    os.environ["ROBOT_DESCRIPTION_URI_PATH"] = (
+        robot_descriptions.ergocub_description.PACKAGE_PATH
+    )
```
flferretti, line 189: Are we still using this?
```suggestion
```

diegoferigo:
Reminder to myself: bump the version of the ROD dependency before merging this PR.

flferretti:
@lorycontixd if you see commits pushed by a bot, don't worry. We can squash those. This should stop as soon as #197 gets merged

lorycontixd:
I was wondering 😂 Thanks for letting me know!

src/jaxsim/parsers/descriptions/collision.py
```diff
@@ -154,3 +154,23 @@ def __eq__(self, other: BoxCollision) -> bool:
...
+
+
+@dataclasses.dataclass
+class MeshCollision(CollisionShape):
+    center: jtp.VectorLike
+
+    def __hash__(self) -> int:
+        return hash(
+            (
+                hash(super()),
```
flferretti, line 166: I believe this is not necessary
```suggestion
```

src/jaxsim/parsers/rod/utils.py
```diff
@@ -202,3 +215,52 @@ def fibonacci_sphere(samples: int) -> npt.NDArray:
...
+        case MeshMappingMethods.RandomSurfaceSampling:
+            points = mesh.sample(nsamples)
+        case MeshMappingMethods.UniformSurfaceSampling:
+            points = trimesh.sample.sample_surface_even(mesh=mesh, count=nsamples)
+        case _:
+            raise ValueError("Invalid mesh mapping method")
+
+    points = mesh.vertices
+    H = collision.pose.transform() if collision.pose is not None else np.eye(4)
+    center_of_collision_wrt_link = (H @ np.hstack([0, 0, 0, 1.0]))[0:-1]
```
flferretti, line 252:
```suggestion
    center_of_collision_wrt_link = H[:3,3]
```

src/jaxsim/parsers/rod/utils.py
```diff
@@ -202,3 +215,52 @@ def fibonacci_sphere(samples: int) -> npt.NDArray:
...
+    mesh_points_wrt_link = (
+        H @ np.hstack([points, np.vstack([1.0] * points.shape[0])]).T
+    )[0:3, :]
```
flferretti, line 255: Why do you need to stack an array of ones to the right of `points`?
```suggestion
    mesh_points_wrt_link = (
        H @ np.hstack([points, np.vstack([1.0] * points.shape[0])]).T
    )[0:3, :]
```

src/jaxsim/parsers/rod/utils.py
```diff
@@ -202,3 +215,52 @@ def fibonacci_sphere(samples: int) -> npt.NDArray:
...
+    collidable_points = [
+        descriptions.CollidablePoint(
+            parent_link=link_description,
+            position=point,
+            enabled=True,
+        )
+        for point in mesh_points_wrt_link.T
```
flferretti, line 262: It would be better to move the transpose operation to the line in which you declare `mesh_points_wrt_link`
```suggestion
        for point in mesh_points_wrt_link
```

flferretti (changes requested):

lorycontixd:
Must find a way to inject the wrapping method from outside the api (`create_mesh_collision` func.) for each mesh, if needed.

CarlottaSartore:
@lorycontixd  @flferretti  can you please check this together? thanks!

lorycontixd:
From the [last commit](5b5550f2838d0bf12bf4edfabee3ae49c72db2a8), the mesh wrapping algorithms don't follow a class inheritance method anymore, but are defined as simple callables. These may be called from ```utils.create_mesh_collision()``` inside the parser which is currently not exposed to the api for the user to modify a mesh.
By default, the mesh is wrapper by collision points along all vertices which is the most simple, accurate method but may become computationally expensive for large meshes.

As far as I know, there could be a few ways to achieve this behaviour:

### Modify jaxsim api
Modify the api from the rod parser straight to the user interface on ```jaxsim.model.JaxSimModel.build_from_model_description``` to allow the user to pass a mesh wrapping algorithm.
An idea could be:

```python
    model_description=urdf_string,
    model_name="passive_walker",
    terrain=terrain,
    is_urdf=True,
    mesh_wrappers={
        "left_foot" : extract_points_uniform_sampling(...),
    }
)
```

##### Advantages: 
- allow the user to modify the mesh to liking straight from the main script
- most generic & user-intuitive method

##### Disadvantages
- must modify numerous function to get it done

### Move logic to rod
Move the mesh wrapping algorithms to [rod](https://github.com/ami-iit/rod) as this feature matches the library's goals of building and modifying robot models.

##### Advantages
?

##### Disadvantages
?

### Leave a default wrapping method
Don't allow the user to modify the mesh wrapping method

##### Advantages
- no more implementation needed

##### Disadvantages
- no control over collision point creation. collision detection can be slow with large meshes
- which method as default?

lorycontixd:
For now choosing the 3rd method, which is to leave a default wrapping algorithm for all meshes. For the algorithm, I'd put a collision point on each vertex of the mesh which is probably the most complete method of all despite its computational load for larger meshes.

flferretti:
> Modify jaxsim api

This is doable, yet to me it looks like an overkill. We should use one single and efficient algorithm for sampling the collidable points, any in-depth modification of the mesh should be done in a different framework and JaxSim is not meant to do that.

> Move logic to rod

This is also doable, but in the same fashion of JaxSim, ROD is not meant to be an in-depth mesh editor.

> Leave a default wrapping method

This looks like the best option for the time being. and regarding
> - [...] collision detection can be slow with large meshes

This can be considered to be tracked by #283. A smart collidables filtering algorithm like GJK is one of the most effective solution in this case.

tests/test_meshes.py
```diff
@@ -0,0 +1,125 @@
...
+
+
+def test_mesh_wrapping_object_mapping():
+    """Test the object mapping method on different meshes.
+    1. Subtract a box from a sphere
+    2. Subtract a sphere from a bigger sphere
+    3. Subtract a small box from a bigger box to remove the right-top corner of the first box
+    """
+
+    return  # Skip this test for now
```
flferretti, line 102: Can we enable this?

tests/test_meshes.py
flferretti, line 1: Can you please add a test in a similar fashion of `test_simulations::test_simulation_with_soft_contacts` with a mesh box?

xela-95:
> > Modify jaxsim api
> 
> This is doable, yet to me it looks like an overkill. We should use one single and efficient algorithm for sampling the collidable points, any in-depth modification of the mesh should be done in a different framework and JaxSim is not meant to do that.
> 
> > Move logic to rod
> 
> This is also doable, but in the same fashion of JaxSim, ROD is not meant to be an in-depth mesh editor.
> 
> > Leave a default wrapping method
> 
> This looks like the best option for the time being. and regarding
> 
> > * [...] collision detection can be slow with large meshes
> 
> This can be considered to be tracked by #283. A smart collidables filtering algorithm like GJK is one of the most effective solution in this case.
> 
> WDYT @xela-95 ?

@flferretti I still have to delve into the details of this PR, but from the latest discussions I've read I fully agree with your comments.

src/jaxsim/parsers/rod/meshes.py
```diff
@@ -0,0 +1,194 @@
...
+        points = sorted_arr[:n]
+    elif direction == "higher":
+        # Select last N points
+        points = sorted_arr[-n:]
+    else:
+        raise ValueError(
+            f"Invalid direction {direction} for SelectPointsOverAxis method"
+        )
+
+    return points
```
flferretti, line 119: As a general rule, it's better to avoid chained conditionals for performance, and for that hashtables (dictionaries) are great
```suggestion
    dirs = {"higher": np.s_[:n], "lower": np.s_[-n:]}

    arr = mesh.vertices

    index = dict(zip(("x", "y", "z"), np.arange(3)))
    
    # Sort the array in ascending order
    arr.sort(axis=index[axis])

    return sorted_arr[dirs[direction]]
```

src/jaxsim/parsers/rod/meshes.py
```diff
@@ -0,0 +1,194 @@
...
+        AssertionError: If the lower bound is greater than the upper bound.
+    """
+
+    # Check bounds
+    upper = upper if upper is not None else np.inf
+    lower = lower if lower is not None else -np.inf
+    assert lower < upper, "Invalid bounds for axis-aligned plane"
+
+    # Logic
+    valid_axes = {"x": 0, "y": 1, "z": 2}
```
flferretti, line 150: At this point, probably `valid_axes` can be global for the module as it is used everywhere

src/jaxsim/parsers/rod/meshes.py
```diff
@@ -0,0 +1,194 @@
...
+
+    Raises:
+        ValueError: If an invalid method is provided.
+    """
+
+    valid_methods = ["subtract", "intersect"]
+    if method not in valid_methods:
+        raise ValueError(f"Invalid method {method} for object mapping")
+    if len(objs) == 0:
+        return mesh.vertices
```
flferretti, line 184: This will correctly return an error in the following lines. The behavior can be confusing if an empty sequence is passed and the user gets no errors

src/jaxsim/parsers/rod/meshes.py
```diff
@@ -0,0 +1,194 @@
...
+    Returns:
+        The extracted points (N x 3 array).
+
+    Raises:
+        ValueError: If an invalid method is provided.
+    """
+
+    valid_methods = ["subtract", "intersect"]
+    if method not in valid_methods:
+        raise ValueError(f"Invalid method {method} for object mapping")
```
flferretti, line 182: Try to use a dictionary as above, this case will be automatically covered by a KeyError

src/jaxsim/parsers/rod/utils.py
flferretti: Unresolving

src/jaxsim/parsers/rod/parser.py
flferretti: Unresolving

tests/test_meshes.py
```diff
@@ -0,0 +1,125 @@
...
+    assert len(points) < len(mesh.vertices)
+
+
+def test_mesh_wrapping_points_over_axis():
+    """Test the points over axis method on different meshes.
+    1. A simple box
+        1.1: Select 10 points from the lower end of the x-axis
+        1.2: Select 10 points from the higher end of the y-axis
+    2. A sphere
+    """
```
flferretti, line 66:
```suggestion
    """
    Test the points over axis method on different meshes.
    1. A simple box
        1.1: Select 10 points from the lower end of the x-axis
        1.2: Select 10 points from the higher end of the y-axis
    2. A sphere
    """
```

flferretti (changes requested):

src/jaxsim/parsers/rod/parser.py
lorycontixd: Removed None check

src/jaxsim/parsers/rod/utils.py
lorycontixd: Changed warning to exception

pyproject.toml
```diff
@@ -68,6 +70,8 @@ testing = [
     "pytest >=6.0",
     "pytest-icdiff",
     "robot-descriptions",
+    "networkx",
+    "lark"
```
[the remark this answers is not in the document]
lorycontixd: Removed dependencies

tests/test_meshes.py
lorycontixd: Done

tests/test_meshes.py
```diff
@@ -0,0 +1,97 @@
...
+        1.2: Remove all points below y=0.0
+    2. A sphere
+    """
+
+    # Test 1.1: Remove all points above x=0.0
+    # The expected result is that the number of points is halved
+    # First, create a box with origin at (0,0,0) and extents (3,3,3) -> points span from -1.5 to 1.5 on axis
+    mesh = trimesh.creation.box(
+        extents=[3.0, 3.0, 3.0],
+    )
```
flferretti, line 40:
```suggestion
    mesh = trimesh.creation.box(extents=[3.0, 3.0, 3.0])
```

tests/test_meshes.py
```diff
@@ -0,0 +1,97 @@
...
+    """Test the AAP wrapping method on different meshes.
+    1. A simple box
+        1.1: Remove all points above x=0.0
+        1.2: Remove all points below y=0.0
+    2. A sphere
+    """
+
+    # Test 1.1: Remove all points above x=0.0
+    # The expected result is that the number of points is halved
+    # First, create a box with origin at (0,0,0) and extents (3,3,3) -> points span from -1.5 to 1.5 on axis
```
flferretti, line 37:
```suggestion
    # Test 1.1: Remove all points above x=0.0.
    #     The expected result is that the number of points is halved.
    #     First, create a box with origin at (0,0,0) and extents (3,3,3), 
    #     i.e. points span from -1.5 to 1.5 on the axis.
```

tests/test_meshes.py
```diff
@@ -0,0 +1,97 @@
...
+    mesh = trimesh.creation.box(
+        extents=[3.0, 3.0, 3.0],
+    )
+    points = meshes.extract_points_aap(mesh=mesh, axis="x", lower=0.0)
+    assert len(points) == len(mesh.vertices) // 2
+    assert all(points[:, 0] > 0.0)
+
+    # Test 1.2: Remove all points below y=0.0
+    # Again, the expected result is that the number of points is halved
```
flferretti, line 46:
```suggestion
    # Test 1.2: Remove all points below y=0.0.
    #     The expected result is that the number of points is halved.
```

tests/test_meshes.py
```diff
@@ -0,0 +1,97 @@
...
+    points = meshes.extract_points_aap(mesh=mesh, axis="y", upper=0.0)
+    assert len(points) == len(mesh.vertices) // 2
+    assert all(points[:, 1] < 0.0)
+
+    # Test 2: A sphere
+    # The sphere is centered at the origin and has a radius of 1.0. Points are expected to be halved
```
flferretti, line 52:
```suggestion
    # Test 2: A sphere.
    #     The sphere is centered at the origin and has a radius of 1.0. 
    #     Points are expected to be halved.
```

tests/test_meshes.py
```diff
@@ -0,0 +1,97 @@
...
+    """
+    Test the points over axis method on different meshes.
+    1. A simple box
+        1.1: Select 10 points from the lower end of the x-axis
+        1.2: Select 10 points from the higher end of the y-axis
+    2. A sphere
+    """
+
+    # Test 1.1: Remove 10 points from the lower end of the x-axis
+    # First, create a box with origin at (0,0,0) and extents (3,3,3) -> points span from -1.5 to 1.5 on axis
```
flferretti, line 70:
```suggestion
    # Test 1.1: Remove 10 points from the lower end of the x-axis.
    #     First, create a box with origin at (0,0,0) and extents (3,3,3),
    #     i.e. points span from -1.5 to 1.5 on the axis.
```

tests/test_meshes.py
```diff
@@ -0,0 +1,97 @@
...
+    mesh = trimesh.creation.box(
+        extents=[3.0, 3.0, 3.0],
+    )
+    points = meshes.extract_points_select_points_over_axis(
+        mesh=mesh, axis="x", direction="lower", n=4
+    )
+    assert len(points) == 4
+    assert all(points[:, 0] < 0.0)
+
+    # Test 1.2: Select 10 points from the higher end of the y-axis
```
flferretti, line 80:
```suggestion
    # Test 1.2: Select 10 points from the higher end of the y-axis.
```

flferretti (commented):
Minor formatting fixes. You can click on `Add suggestion to batch` and commit the suggestions all together

tests/test_meshes.py
flferretti: After discussing f2f, we decided to add this test in a future release. I opened https://github.com/ami-iit/jaxsim/issues/289 for that

flferretti (approved):
Thanks a lot @lorycontixd, great job! LGTM

PR #157, opened 2024-05-21
Fix method on `jaxsim.math.Transform`

flferretti:
This pull request fixes a method on the `jaxsim.math.Transform` class. The method `from_rotation_and_translation` was updated to correctly use the `jaxlie.SO3` type for the rotation. This ensures that the method `jaxlie.SE3.from_rotation_and_translation` gets called with the correct input types

----
📚 Documentation preview 📚: https://jaxsim--157.org.readthedocs.build//157/


diegoferigo (approved):
Great, thanks for the fix! This method is never used in the codebase nor tested.

PR #148, opened 2024-05-07
Expose quantities related to generic frames

xela-95:
Closes #147 

----
📚 Documentation preview 📚: https://jaxsim--148.org.readthedocs.build//148/

PR #158, opened 2024-05-21
Fix processing of parsed frames

diegoferigo:
Completes #150 and provides the last necessary ingredients for #148.

### DIFF
--- a/src/jaxsim/parsers/descriptions/model.py
+++ b/src/jaxsim/parsers/descriptions/model.py
@@ -115,7 +115,7 @@ class ModelDescription(KinematicGraph):
             final_collisions.append(new_collision_shape)
 
             # If the frame was found, update the collidable points' pose and add them
-            # to the new collision shape
+            # to the new collision shape.
             for cp in collision_shape.collidable_points:
                 # Find the link that is part of the (reduced) model in which the
                 # collision shape's parent was lumped into
--- a/src/jaxsim/parsers/descriptions/model.py
+++ b/src/jaxsim/parsers/descriptions/model.py
@@ -35,6 +35,7 @@ class ModelDescription(KinematicGraph):
         name: str,
         links: List[LinkDescription],
         joints: List[JointDescription],
+        frames: List[LinkDescription] | None = None,
         collisions: List[CollisionShape] = (),
         fixed_base: bool = False,
         base_link_name: str | None = None,
@@ -65,6 +66,7 @@ class ModelDescription(KinematicGraph):
         kinematic_graph = KinematicGraph.build_from(
             links=links,
             joints=joints,
+            frames=frames,
             root_link_name=base_link_name,
             root_pose=model_pose,
         )
@@ -172,6 +174,7 @@ class ModelDescription(KinematicGraph):
             name=self.name,
             links=list(self.links_dict.values()),
             joints=self.joints,
+            frames=self.frames,
             collisions=self.collision_shapes,
             fixed_base=self.fixed_base,
             base_link_name=list(iter(self))[0].name,
--- a/src/jaxsim/parsers/descriptions/model.py
+++ b/src/jaxsim/parsers/descriptions/model.py
@@ -1,6 +1,8 @@
+from __future__ import annotations
+
 import dataclasses
 import itertools
-from typing import List
+from typing import Sequence
 
 from jaxsim import logging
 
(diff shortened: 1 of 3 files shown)


src/jaxsim/parsers/kinematic_graph.py
```diff
@@ -776,3 +881,30 @@ def pre_H_suc(
...
+
+        match frame.parent.name:
+            case parent_name if parent_name in self.graph.links_dict:
+                return parent_name
+
+            case parent_name if parent_name in self.graph.frames_dict:
+                return self.find_parent_link_of_frame(name=parent_name)
+
+            case _:
+                msg = f"Failed to find element with name '{frame.parent.name}'"
```
flferretti, line 909: This looks a bit ambiguous to me, can we be a bit more specific?
```suggestion
                msg = f"Failed to find parent element of frame '{name}' with name '{frame.parent.name}'"
```

src/jaxsim/parsers/kinematic_graph.py
```diff
@@ -452,58 +533,77 @@ def reduce(self, considered_joints: List[str]) -> KinematicGraph:
...
+            # Always find the real parent link of the frame
+            name_of_new_parent_link = fk_reduced.find_parent_link_of_frame(
+                name=frame.name
             )
+            assert name_of_new_parent_link in reduced_graph, name_of_new_parent_link
+
+            # Notify the user if the parent link has changed.
+            if name_of_new_parent_link != frame.parent.name:
+                msg = "New parent of frame '{}' is '{}'"
+                logging.debug(msg.format(frame.name, name_of_new_parent_link))
```
flferretti, line 552: Minor issue, but it's to be coherent with the library style
```suggestion
                logging.debug(msg=msg.format(frame.name, name_of_new_parent_link))
```

src/jaxsim/parsers/kinematic_graph.py
```diff
@@ -427,19 +496,31 @@ def reduce(self, considered_joints: List[str]) -> KinematicGraph:
...
+            root_link_name=full_graph.root.name,
+        )
+
+        assert set(f.name for f in self.frames).isdisjoint(
+            set(f.name for f in unconnected_frames + reduced_frames)
         )
 
-        # Create the reduced graph
+        for link in unconnected_links:
+            logging.debug(msg=f"Link '{link.name}' is unconnected and became a frame")
```
flferretti, line 517: Definitely not an issue, but I believe that if we use an `f-string` here, Python will try to evaluate this even when the logging level is above `debug`
```suggestion
            logging.debug(msg="Link '{}' is unconnected and became a frame".format(link.name))
```

src/jaxsim/parsers/descriptions/model.py
```diff
@@ -112,7 +115,7 @@ def build_model_from(
             final_collisions.append(new_collision_shape)
 
             # If the frame was found, update the collidable points' pose and add them
-            #  to the new collision shape
+            # to the new collision shape
```
flferretti, line 118: Just if you want to be coherent with the rest of the comments style
```suggestion
            # to the new collision shape.
```

flferretti (approved):
Thanks a lot Diego! LGTM, I just left some minor comments

src/jaxsim/parsers/kinematic_graph.py
diegoferigo: Mmh I think that the string is evaluated regardless since it's not controlled by the logger. It's only the redirection to stdout that is filtered by the logger, all the processing should happen in any case (never checked, just my intuition).

The situation would be different if the fmt processing would be finalized internally like it happens in jax debug messages (for other reasons, in that case).

flferretti: In this case it's definitely not crucial as we are not evaluating anything computationally expensive, so you can ignore this comment. See https://docs.astral.sh/ruff/rules/logging-f-string/ for reference

diegoferigo: Good to know, thanks for the reference. Btw, what you aimed is not currently possible with our wrapper class of the python logger. The evaluation can be deferred by using the plain logger as:

```python
logger.debug(msg="My string printing {something}", something=something)
```

However, out logger only exposes `msg` and not the optional args /  kwargs.

PR #161, opened 2024-05-27
Add support for computing the bias acceleration of the center of mass

diegoferigo:
> [!WARNING]  
> WIP


diegoferigo:
I've narrowed down the failures in CI. The new test is currently failing when:

- The model is floating base.
- The velocity representation is Mixed.
- Then base angular velocity $\boldsymbol{\omega}_{W,B}$ is not zero.

diegoferigo:
After chatting with @traversaro, we realized that the previous logic was setting to zero the wrong $\dot{\boldsymbol{\nu}}$. The solution is to compute the bias link acceleration in the active representation (implemented in #127, that internally sets to zero the right $\dot{\boldsymbol{\nu}}$), and then convert back these accelerations to body-fixed representation to obtain the bias ${}_G \dot{\mathbf{h}}$.

I believe that we can unify somhow the complicated logic used in the function to compute the bias accelerations of links and CoM, but this can be done in another PR. I want to move this forward since it is finally working fine.

flferretti (approved):
Thanks a lot Diego, LGTM!

PR #162, opened 2024-05-28
Ensure consistent link forces and fix static joint friction

flferretti:
This pull request includes several changes to ensure consistent link forces and fix the base acceleration transform in the code. The changes include:

- Ensuring that link forces are in `VelRepr.Inertial` when passed to forward dynamics
- Fixing the base acceleration transform in `VelRepr.Mixed` in ABA
- Using joint velocities sign for static joint friction
- Adding a test to verify that linear forces are applied correctly

C.C. @xela-95


diegoferigo (approved):
Awesome! Thanks for spotting these two bugs (static joint friction and missing conversion of link forces passed by the user).

PR #163, opened 2024-05-31
Compute the pose and the jacobian of the implicit frames associated to collidable points

diegoferigo:
We consider the following implicit frame associated to each collidable point:

```math
C := \left({}^W \mathbf{p}_C, [L]\right)
```

where:

- ${}^W \mathbf{p}_C \in \mathbb{R}^3$ is the position of the collidable point in world coordinates.
- $[L]$ is the orientation frame of the parent link of the collidable point (i.e. the link it is rigidly attached to).

This PR implements the following:

- The world-to-contact-frame transform ${}^W \mathbf{H}_C \in \text{SE}(3)$.
- The free-floating jacobian ${}^O J_{W,C/I}$ of the contact frame, where $I$ is the input representation that should match the representation of $\boldsymbol{\nu} \in \mathbb{R}^{6+n}$, and $O$ the desired output representation.

----
📚 Documentation preview 📚: https://jaxsim--163.org.readthedocs.build//163/

### DIFF
--- a/src/jaxsim/api/contact.py
+++ b/src/jaxsim/api/contact.py
@@ -369,8 +369,8 @@ def jacobian(
                 C_X_W = jaxsim.math.Adjoint.from_transform(
                     transform=W_H_C, inverse=True
                 )
-                C_J_WCi = C_X_W @ W_J_WC
-                return C_J_WCi
+                C_J_WC = C_X_W @ W_J_WC
+                return C_J_WC
 
             O_J_WC = jax.vmap(jacobian)(W_H_C, W_J_WC)
 
--- a/src/jaxsim/api/contact.py
+++ b/src/jaxsim/api/contact.py
@@ -306,3 +306,87 @@ def transforms(model: js.model.JaxSimModel, data: js.data.JaxSimModelData) -> jt
     # Compose the work-to-link and link-to-point transforms.
     return jax.vmap(lambda W_H_Li, L_H_Ci: W_H_Li @ L_H_Ci)(W_H_L, L_H_C)
 
+
+@functools.partial(jax.jit, static_argnames=["output_vel_repr"])
+def jacobian(
+    model: js.model.JaxSimModel,
+    data: js.data.JaxSimModelData,
+    *,
+    output_vel_repr: VelRepr | None = None,
+) -> jtp.Array:
+    r"""
+    Return the free-floating Jacobian of the collidable points.
+
+    Args:
+        model: The model to consider.
+        data: The data of the considered model.
+        output_vel_repr:
+            The output velocity representation of the free-floating jacobian.
+
+    Returns:
+        The stacked 6×(6+n) free-floating jacobians of the frames associated to the
+        collidable points.
+
+    Note:
+        Each collidable point is implicitly associated with a frame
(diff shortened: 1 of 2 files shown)


tests/test_contact.py
```diff
@@ -0,0 +1,37 @@
...
+    # =====
+
+    # Compute the velocity of the collidable points with a RBDA.
+    # This function always returns the linear part of the mixed velocity of the
+    # implicit frame C corresponding to the collidable point.
+    W_ṗ_C = js.contact.collidable_point_velocities(model=model, data=data)
+
+    # Compute the generalized velocity and the free-floating Jacobian of the frame C.
+    ν = data.generalized_velocity()
+    CL_J_WC = js.contact.jacobian(model=model, data=data, output_vel_repr=VelRepr.Mixed)
```
xela-95, line 32: Is this correct? I was expecting it to be ${} ^{C[W]} J _{W, C}$ 🤔

diegoferigo: Yep you're right, that's the mixed velocity and I should have named the variable `CW_J_WC`. Thanks!

xela-95 (approved):

traversaro (approved):

src/jaxsim/api/contact.py
```diff
@@ -269,3 +274,124 @@ def estimate_model_height(model: js.model.JaxSimModel) -> jtp.Float:
     )
 
     return sc_parameters
+
+
+@jax.jit
+def transforms(model: js.model.JaxSimModel, data: js.data.JaxSimModelData) -> jtp.Array:
+    r"""
+    Return the pose of the collidable points.
```
flferretti, line 282: Can we be a little bit more verbose?

```suggestion
    Return the pose of the collidable points in inertial-fixed representation.
```

src/jaxsim/api/contact.py
```diff
@@ -269,3 +274,124 @@ def estimate_model_height(model: js.model.JaxSimModel) -> jtp.Float:
...
+        case VelRepr.Body:
+
+            W_H_C = transforms(model=model, data=data)
+
+            def jacobian(W_H_C: jtp.Matrix, W_J_WC: jtp.Matrix) -> jtp.Matrix:
+                C_X_W = jaxsim.math.Adjoint.from_transform(
+                    transform=W_H_C, inverse=True
+                )
+                C_J_WCi = C_X_W @ W_J_WC
+                return C_J_WCi
```
flferretti, line 373: Just to be consistent with the notation used below

```suggestion
                C_J_WC = C_X_W @ W_J_WC
                return C_J_WC
```

flferretti (approved):
Great! Thanks Diego, LGTM

src/jaxsim/api/contact.py
diegoferigo: The $\text{SE}(3)$ pose does not depend on any velocity representation. In JaxSim, poses are always absolute, i.e. w.r.t. the world frame $W$.

src/jaxsim/api/contact.py
diegoferigo: Yep I changed the notation in a later stage of this PR, and this is a leftover of the old one, thanks!

PR #165, opened 2024-06-03
Fix passing different PyTrees to JIT-compiled functions

diegoferigo:
This PR  should finally solve #103 also for models with joints, in particular:

- Properly implements the `__eq__` magic method of all classes that are used in JIT compiled functions.
- Introduces a new class `jaxsim.utils.wrappers.HashedNumpyArray` that can wrap NumPy and JAX NumPy array to provide them a hash and equality operators. This is necessary for all `jax_dataclasses.Static` fields storing arrays.
- Updates the pattern to hide the objects wrapped with `HashedNumpyArray` and `HashlessObject` exposing only the underlying object through a class property.
- Fixed `JointModel.joint_axis` attribute to properly contain `JointGenericAxis` objects. Before this PR (from [#153](https://github.com/ami-iit/jaxsim/pull/153) I guess), this attribute was typed to hold a `JointGenericAxis` but instead it stored plain JAX numpy arrays with the 3D vector of the axis.

----
📚 Documentation preview 📚: https://jaxsim--165.org.readthedocs.build//165/


diegoferigo:
This PR also fixes the M(N)WE provided by @flferretti in https://github.com/ami-iit/jaxsim/issues/103#issuecomment-2058752595.

src/jaxsim/api/data.py
```diff
@@ -43,6 +43,24 @@ class JaxSimModelData(common.ModelDataWithVelocityRepresentation):
...
     )
 
+    def __hash__(self) -> int:
+
+        return hash(
+            (
+                hash(self.state),
+                hash(tuple(self.gravity.flatten().tolist())),
+                hash(self.soft_contacts_params),
+                hash(jnp.atleast_1d(self.time_ns).flatten().tolist()),
```
flferretti, line 53: I guess that `time_ns` has shape `()`, so probably a `.tolist()` of a `.item()` should be enough
```suggestion
                hash(self.time_ns.tolist()),
```

src/jaxsim/api/data.py
```diff
@@ -43,6 +43,24 @@ class JaxSimModelData(common.ModelDataWithVelocityRepresentation):
         default_factory=lambda: jnp.array(0, dtype=jnp.uint64)
     )
 
+    def __hash__(self) -> int:
+
+        return hash(
+            (
+                hash(self.state),
+                hash(tuple(self.gravity.flatten().tolist())),
```
flferretti, line 51: `self.gravity` has shape `(3,)`, so probably the `.flatten()` is not necessary
```suggestion
                hash(tuple(self.gravity.tolist())),
```

src/jaxsim/api/model.py
```diff
@@ -260,7 +279,7 @@ def frame_names(self) -> tuple[str, ...]:
             The names of the links in the model.
         """
 
-        return tuple([frame.name for frame in self.description.get().frames])
+        return tuple([frame.name for frame in self.description.frames])
```
flferretti, line 282: We could use a generator here
```suggestion
        return tuple(frame.name for frame in self.description.frames)
```

src/jaxsim/api/ode_data.py
```diff
@@ -281,6 +281,24 @@ class PhysicsModelState(JaxsimDataclass):
...
     )
 
+    def __hash__(self) -> int:
+
+        return hash(
+            (
+                hash(tuple(jnp.atleast_1d(self.joint_positions.flatten().tolist()))),
+                hash(tuple(jnp.atleast_1d(self.joint_velocities.flatten().tolist()))),
+                hash(tuple(self.base_position.flatten().tolist())),
+                hash(tuple(self.base_quaternion.flatten().tolist())),
```
flferretti, line 291: Probably I'm missing something, is the `.flatten()` necessary here?
```suggestion
                hash(tuple(jnp.atleast_1d(self.joint_positions.tolist()))),
                hash(tuple(jnp.atleast_1d(self.joint_velocities.tolist()))),
                hash(tuple(self.base_position.tolist())),
                hash(tuple(self.base_quaternion.tolist())),
```

src/jaxsim/rbda/soft_contacts.py
```diff
@@ -29,6 +29,23 @@ class SoftContactsParams(JaxsimDataclass):
         default_factory=lambda: jnp.array(0.5, dtype=float)
     )
 
+    def __hash__(self) -> int:
+
+        return hash(
+            (
+                hash(tuple(jnp.atleast_1d(self.K).flatten().tolist())),
+                hash(tuple(jnp.atleast_1d(self.D).flatten().tolist())),
+                hash(tuple(jnp.atleast_1d(self.mu).flatten().tolist())),
```
flferretti, line 38: Same as above, is the `.flatten()` necessary?
```suggestion
                hash(tuple(jnp.atleast_1d(self.K).tolist())),
                hash(tuple(jnp.atleast_1d(self.D).tolist())),
                hash(tuple(jnp.atleast_1d(self.mu).tolist())),
```

tests/conftest.py
```diff
@@ -177,18 +177,19 @@ def jaxsim_model_sphere() -> js.model.JaxSimModel:
 
 
 @pytest.fixture(scope="session")
```
flferretti, line 179: Is this still necessary now?
```suggestion
```

flferretti (approved):
Thanks Diego, LGTM!

src/jaxsim/api/data.py
diegoferigo: The operator `jnp.atleast1d(X).flatten()` is there mainly for making the hash function work on vectorized pytrees.

src/jaxsim/api/data.py
diegoferigo: The operator `jnp.atleast1d(X).flatten()` is there mainly for making the hash function work on vectorized pytrees.

src/jaxsim/api/ode_data.py
diegoferigo: The operator `jnp.atleast1d(X).flatten()` is there mainly for making the hash function work on vectorized pytrees.

tests/conftest.py
diegoferigo: This can be computed only once and cached for the entire pytest session. It is not necessary but why not caching it.

src/jaxsim/rbda/soft_contacts.py
diegoferigo: The operator `jnp.atleast1d(X).flatten()` is there mainly for making the hash function work on vectorized pytrees.

tests/conftest.py
flferretti: Alright, thanks for the explanation

PR #167, opened 2024-06-04
Always use the link frame in body-fixed and mixed output representations of model Jacobian

diegoferigo:


traversaro (approved):

flferretti (approved):
LGTM, thanks!

PR #170, opened 2024-06-05
Fix prismatic joints transform in `math.joint_model.supported_joint_motion`

flferretti:
The method `jaxlie.SE3.from_translation` used to calculate the parent-child transform for prismatic joints doesn't actually exists

----
📚 Documentation preview 📚: https://jaxsim--170.org.readthedocs.build//170/


diegoferigo (approved):
My bad, I think here Copilot tricked me and we don't have checks with models having prismatic joints (not a priority). Thanks!

PR #171, opened 2024-06-10
Test J̇ computation against AD

diegoferigo:
This PR extends the test included in #169 to compare the computation of $\dot{J}$ with the corresponding derivative computed as:

```math
\dot{J}(\mathbf{q},\, \boldsymbol{\nu}) = \frac{\text{d} J(\mathbf{q})}{\text{d} t} = \frac{\partial J}{\partial \mathbf{q}} \frac{\partial \mathbf{q}}{\partial t} = \frac{\partial J}{\partial \mathbf{q}} \dot{\mathbf{q}}
```

As usual, it's worth noting that since $\mathbf{q} = ( {}^W \mathbf{p}_B \\,; {}^W \mathtt{Q}_B\\,; \mathbf{s} ) \in \mathbb{R}^{7+n}$, we have that $\dot{\mathbf{q}} \neq \boldsymbol{\nu}$. However, by taking extra care of the derivative of the base quaternion, we can compute $\dot{\mathbf{q}}$ and use it in the test (that, btw, is independent from the velocity representation).

Triggered by https://github.com/ami-iit/jaxsim/pull/169#issuecomment-2156708169 from @DanielePucci.

----
📚 Documentation preview 📚: https://jaxsim--171.org.readthedocs.build//171/

### DIFF
--- a/tests/test_api_link.py
+++ b/tests/test_api_link.py
@@ -3,6 +3,7 @@ import jax.numpy as jnp
 import pytest
 
 import jaxsim.api as js
+import jaxsim.math
 from jaxsim import VelRepr
 
 from . import utils_idyntree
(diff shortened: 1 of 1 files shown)


diegoferigo:
Out of curiosity, I've tried to benchmark on CPU the two implementations:

|| Analytical $\dot{J}$ | AD $\dot{J}$ |
|:--- | :---: | :---: |
| JIT compilation| 2.08 s | 4.04 s |
| Runtime | 248 µs ± 18.5 µs | 6.65 ms ± 633 µs |

This numbers refer to the computation of $\dot{J}$ of a single link of the ErgoCub robot (57 DoFs, pretty large). It seems pretty clear that the analytical computation has always to be preferred.

diegoferigo:
This could be interesting to all @ami-iit/vertical_control-oriented-learning.

traversaro (approved):

tests/test_api_link.py
```diff
@@ -250,4 +251,66 @@ def test_link_jacobian_derivative(
...
+    q̇ = compute_q̇(data)
+
+    # Compute dJ/dt with AD.
+    dJ_dq = jax.jacfwd(J, argnums=0)(q)
+    O_J̇_ad_WL_I = jnp.einsum("ijkq,q->ijk", dJ_dq, q̇)
+
+    assert O_J̇_ad_WL_I == pytest.approx(O_J̇_WL_I)
+    assert jnp.einsum("l6g,g->l6", O_J̇_ad_WL_I, I_ν) == pytest.approx(
+        jnp.einsum("l6g,g->l6", O_J̇_WL_I, I_ν)
+    )
```
flferretti, line 316: I think I'm missing something, can we just compare the two ${} ^O\dot{J} _{WL, I}$?
```suggestion
    assert O_J̇_ad_WL_I == pytest.approx(O_J̇_WL_I)
```

flferretti (approved):
This is great! Thanks Diego

tests/test_api_link.py
diegoferigo: The problem is that the elements of $\dot{J}$ can be really close to $0$. Computing the link bias acceleration, instead, provides larger values since they sum up along kinematic chains.

I think it's more robust having also the projected values checked. I fear that comparing the Jacobians with too small numbers may provide a `0 = 0` even if the values are wrong (but so small that are within the default tolerances of pytest). And, I don't want to mess us with the tolerances. If something related to them is wrong, the last assert will fail.

flferretti: Great, thanks for the explanation

PR #174, opened 2024-06-13
Update CI/CD workflow to interface with PyPI as Trusted Publisher

diegoferigo:
More info about Trusted Publishers in PyPI: https://docs.pypi.org/trusted-publishers/.

I've enabled the `ci_cd.yml` workflow in PyPI as follows:

image

----
📚 Documentation preview 📚: https://jaxsim--174.org.readthedocs.build//174/


flferretti (approved):
LGTM, thanks!

PR #178, opened 2024-06-13
Add `ContactModel` base class and abstract contact handling in `JaxSimModel` and `JaxSimModelData`

flferretti:
This pull request refactors the contact models to enhance flexibility and future extensibility. Key changes include:

- Introduced a `ContactModel` base class together with `ContactsState` and `ContactParams`
- Refactored `SoftContactModel`, `SoftContactParams` and `SoftContactState` to inherit from `ContactModel` for consistency.
- Added `contact_model` attribute to `JaxSimModel`.

These changes allow specifying different contact models during instantiation, facilitating the development of additional contact models in the future.

C.C. @xela-95 

----
📚 Documentation preview 📚: https://jaxsim--178.org.readthedocs.build//178/


flferretti:
For some reason, CI is failing only in Python 3.10 (https://github.com/ami-iit/jaxsim/actions/runs/9511311149)

diegoferigo:
> For some reason, CI is failing only in Python 3.10 (https://github.com/ami-iit/jaxsim/actions/runs/9511311149)

Mmh, since it is related to hash and equality, maybe we can wait to merge #179.

src/jaxsim/api/contact.py
```diff
@@ -395,3 +406,96 @@ def mixed_jacobian(W_H_C: jtp.Matrix, W_J_WC: jtp.Matrix) -> jtp.Matrix:
             raise ValueError(output_vel_repr)
 
     return O_J_WC
+
+
+@jax_dataclasses.pytree_dataclass
+class ContactsState(JaxsimDataclass, abc.ABC):
```
diegoferigo, line 412: If possible, I'd keep this a plain class that just inherits from `abc.ABC`, having all methods marked as abstract (they can also be static).

src/jaxsim/api/contact.py
```diff
@@ -395,3 +406,96 @@ def mixed_jacobian(W_H_C: jtp.Matrix, W_J_WC: jtp.Matrix) -> jtp.Matrix:
...
+    def valid(self, **kwargs) -> bool:
+        """
+        Check if the contacts state is valid.
+        """
+
+        return True
+
+
+@jax_dataclasses.pytree_dataclass
+class ContactsParams(JaxsimDataclass, abc.ABC):
```
diegoferigo, line 446: Same comment as for `ContactState`.

src/jaxsim/api/contact.py
```diff
@@ -395,3 +406,96 @@ def mixed_jacobian(W_H_C: jtp.Matrix, W_J_WC: jtp.Matrix) -> jtp.Matrix:
...
+    Abstract class representing a contact model.
+    Attributes:
+        parameters: The parameters of the contact model.
+        terrain: The terrain model.
+    """
+
+    parameters: ContactsParams = dataclasses.field(default_factory=ContactsParams)
+    terrain: jaxsim.terrain.Terrain = dataclasses.field(
+        default_factory=jaxsim.terrain.FlatTerrain
+    )
```
diegoferigo, line 483: Also here, this class can just be a standard class inheriting from `abc.ABC`. The attributes `parameters` and `terrain` could be class attributes. This is a way to mark those attribute as mandatory for child classes (and if child classes are dataclasses, there would be no problem).

src/jaxsim/api/contact.py
```diff
@@ -395,3 +406,96 @@ def mixed_jacobian(W_H_C: jtp.Matrix, W_J_WC: jtp.Matrix) -> jtp.Matrix:
...
+        """
+        Compute the contact forces.
+        Args:
+            position: The position of the collidable point.
+            velocity: The velocity of the collidable point.
+        Returns:
+            A tuple containing the contact force and additional information.
+        """
+
+        raise NotImplementedError
```
diegoferigo, line 501: Usually `abc.abstractmethod` does not require raising exceptions, it alredy raises if downstream classes do no implement the method. A simple `pass` AFAIK is more common.

src/jaxsim/api/ode_data.py
```diff
@@ -116,11 +120,11 @@ class ODEState(JaxsimDataclass):
 
     Attributes:
         physics_model: The state of the physics model.
-        soft_contacts: The state of the soft-contacts model.
+        contacts_state: The state of the contacts model.
     """
 
     physics_model: PhysicsModelState
-    soft_contacts: SoftContactsState
+    contacts_state: js.contact.ContactsState
```
diegoferigo, line 127: Why not just:

```suggestion
    contact: js.contact.ContactsState
```

src/jaxsim/api/ode_data.py
```diff
@@ -158,6 +162,15 @@ def build_from_jaxsim_model(
...
         """
 
+        # Get the contact model from the `JaxSimModel`
+        prefix = type(model.contact_model).__name__.split("Contact")[0]
+
+        if prefix:
+            module_name = f"{prefix.lower()}_contacts"
+            class_name = f"{prefix.capitalize()}ContactsState"
+        else:
+            raise ValueError("Unable to determine contact state class prefix.")
```
diegoferigo, line 172: Would `isinstance` on `model.contact_model` work here? You can import all supported contact model inside this method. In the future, we can think of better ways using multiple dispatch for allowing users to define custom models, if needed. However, I'd keep things simple for now.

src/jaxsim/api/ode_data.py
```diff
@@ -169,24 +182,30 @@ def build_from_jaxsim_model(
                 base_linear_velocity=base_linear_velocity,
                 base_angular_velocity=base_angular_velocity,
             ),
-            soft_contacts_state=SoftContactsState.build_from_jaxsim_model(
+            contacts_state=getattr(
+                importlib.import_module(f"jaxsim.api.{module_name}"), class_name
+            ).build_from_jaxsim_model(
```
diegoferigo, line 187: If you import the supported contact models above as suggested, you can remove this piece of code that, in any case, would not support user-defined contact models (outside `jaxsim`).

src/jaxsim/api/contact.py
```diff
@@ -395,3 +406,96 @@ def mixed_jacobian(W_H_C: jtp.Matrix, W_J_WC: jtp.Matrix) -> jtp.Matrix:
...
+        default_factory=jaxsim.terrain.FlatTerrain
+    )
+
+    @abc.abstractmethod
+    def contact_model(
+        self,
+        position: jtp.Vector,
+        velocity: jtp.Vector,
+        **kwargs,
+    ) -> tuple[jtp.Vector, jtp.Vector]:
```
diegoferigo, line 491: If we abstract the soft contacts class, I'd take advantage for renaming this method as follows:

```suggestion
    @abc.abstractmethod
    def compute_contact_forces(
        self,
        position: jtp.Vector,
        velocity: jtp.Vector,
        **kwargs,
    ) -> tuple[jtp.Vector, tuple[Any, ...]]:
```

Usually, more advanced contact models, operate in different stages. We might need to introduce new methods associated to those in stages in the future for alternative contact models.

diegoferigo (changes requested):
Thanks @flferretti for this first attempt aimed at abstracting the contact model in view of the addition of alternative ones. I left few initial comments in the code.

- What if we leave the actual soft-contact class defining the algorithm in `jaxsim.rbda`? If we plan to add multiple contact models, I'm ok in adding a new subpackage `jaxsim.rbda.contacts.{soft|...}`.
- I'd keep the high-level detection of the contact model much more simple, right now there are many hacks I'd like to avoid at this initial stage.

flferretti:
> What if we leave the actual soft-contact class defining the algorithm in `jaxsim.rbda`? If we plan to add multiple contact models, I'm ok in adding a new subpackage `jaxsim.rbda.contacts.{soft|...}`.

Done in https://github.com/ami-iit/jaxsim/pull/178/commits/8120bdf9d85cfe045d12788f7122e000fc8b4077

src/jaxsim/api/contact.py
flferretti: Done in https://github.com/ami-iit/jaxsim/pull/178/commits/6542f8bb0cbd3d0607ba4db9a909740c8da63951

src/jaxsim/api/contact.py
flferretti: Done in https://github.com/ami-iit/jaxsim/pull/178/commits/6542f8bb0cbd3d0607ba4db9a909740c8da63951

src/jaxsim/api/contact.py
flferretti: Done in https://github.com/ami-iit/jaxsim/pull/178/commits/6542f8bb0cbd3d0607ba4db9a909740c8da63951

src/jaxsim/api/contact.py
flferretti: Done in https://github.com/ami-iit/jaxsim/pull/178/commits/6542f8bb0cbd3d0607ba4db9a909740c8da63951

src/jaxsim/api/contact.py
flferretti: Done in https://github.com/ami-iit/jaxsim/pull/178/commits/d8ef45331383e92253bb310441b0afea1f662c81

src/jaxsim/api/ode_data.py
flferretti: Done in https://github.com/ami-iit/jaxsim/pull/178/commits/4556dbb2b89dbe91b1989903973d93e448e2e8bb

src/jaxsim/api/ode_data.py
flferretti: Done in https://github.com/ami-iit/jaxsim/pull/178/commits/9675c4965004291dd3d78cae1546ff539b9e3f80

src/jaxsim/api/ode_data.py
flferretti: Done in https://github.com/ami-iit/jaxsim/pull/178/commits/9675c4965004291dd3d78cae1546ff539b9e3f80

src/jaxsim/api/ode_data.py
flferretti: Done in https://github.com/ami-iit/jaxsim/pull/178/commits/4556dbb2b89dbe91b1989903973d93e448e2e8bb

diegoferigo (changes requested):
Great, thanks @flferretti! Last review pass with the final suggestions. Well done :rocket:

src/jaxsim/api/contact.py
```diff
@@ -357,12 +365,10 @@ def jacobian(
 
     # Adjust the output representation.
     match output_vel_repr:
-
```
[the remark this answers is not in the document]
flferretti: Sure! A possible idea to avoid this kind of formatting conflicts might be to use `ruff` formatter instead of `black`, which automatically removed these kind of blank lines.

src/jaxsim/rbda/contacts/common.py
```diff
@@ -0,0 +1,97 @@
...
+        raise NotImplementedError
+
+    def valid(self, *args, **kwargs) -> bool:
+        """
+        Check if the parameters are valid.
+        Returns:
+            True if the parameters are valid, False otherwise.
+        """
+
+        return True
```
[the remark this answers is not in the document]
flferretti: The `SoftContactParams` does not implement that explicitly, that's why I avoided marking it as an abstract method, but I can add that

src/jaxsim/api/contact.py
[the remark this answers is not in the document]
diegoferigo: Wait, I want to keep these lines, not removing them :)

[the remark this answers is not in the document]
diegoferigo: Black removes these lines only in a particular -and kind of unfortunate- version, I think either the 2022 or 2023 one. Now I pinned 2024 [project-wise](https://github.com/ami-iit/jaxsim/blob/35f25b1d006a604257684bc9b89c622d8400ccec/setup.cfg#L72).

src/jaxsim/rbda/contacts/common.py
```diff
@@ -0,0 +1,97 @@
...
+        velocity: jtp.Vector,
+        **kwargs,
+    ) -> tuple[Any, ...]:
+        """
+        Compute the contact forces.
+        Args:
+            position: The position of the collidable point.
+            velocity: The velocity of the collidable point.
+        Returns:
+            A tuple containing the contact force and additional information.
```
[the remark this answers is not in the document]
flferretti: Done in https://github.com/ami-iit/jaxsim/pull/178/commits/f64764fa5e02bd76438e18c0fde2c119115297b4

src/jaxsim/rbda/contacts/common.py
```diff
@@ -0,0 +1,97 @@
...
+    parameters: ContactsParams
+    terrain: jaxsim.terrain.Terrain
+
+    @abc.abstractmethod
+    def compute_contact_forces(
+        self,
+        position: jtp.Vector,
+        velocity: jtp.Vector,
+        **kwargs,
+    ) -> tuple[Any, ...]:
```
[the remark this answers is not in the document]
flferretti: Done in https://github.com/ami-iit/jaxsim/pull/178/commits/f64764fa5e02bd76438e18c0fde2c119115297b4

src/jaxsim/rbda/contacts/common.py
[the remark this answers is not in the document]
flferretti: Done in https://github.com/ami-iit/jaxsim/pull/178/commits/f64764fa5e02bd76438e18c0fde2c119115297b4

src/jaxsim/api/contact.py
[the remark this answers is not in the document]
flferretti: Done in

src/jaxsim/rbda/contacts/common.py
```diff
@@ -0,0 +1,97 @@
...
+    @classmethod
+    @abc.abstractmethod
+    def build(cls) -> ContactsParams:
+        """
+        Create a `ContactsParams` instance with specified parameters.
+        Returns:
+            The `ContactsParams` instance.
+        """
+
+        raise NotImplementedError
```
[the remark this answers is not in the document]
flferretti: Done in https://github.com/ami-iit/jaxsim/pull/178/commits/f64764fa5e02bd76438e18c0fde2c119115297b4

src/jaxsim/rbda/contacts/common.py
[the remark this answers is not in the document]
diegoferigo: Note: this has to be propagated to the implementation before resolving this conversation.

src/jaxsim/api/model.py
```diff
@@ -122,11 +129,15 @@ def build_from_model_description(
                 considered_joints=considered_joints
             )
 
+        terrain = terrain or JaxSimModel.__dataclass_fields__["terrain"].default
+        contact_model = contact_model or SoftContacts(terrain=terrain)
+
```
diegoferigo, line 134: Can this initialization be removed since `JaxSimModel.build` already handles the case in which `terrain=None` and `contact_model=None`?

src/jaxsim/rbda/contacts/common.py
```diff
@@ -0,0 +1,95 @@
...
+
+    @classmethod
+    @abc.abstractmethod
+    def build(cls) -> ContactsParams:
+        """
+        Create a `ContactsParams` instance with specified parameters.
+        Returns:
+            The `ContactsParams` instance.
+        """
+
```
diegoferigo, line 55:
```suggestion
        """
        pass
```

src/jaxsim/rbda/contacts/common.py
```diff
@@ -0,0 +1,95 @@
...
+    @abc.abstractmethod
+    def valid(self, *args, **kwargs) -> bool:
+        """
+        Check if the parameters are valid.
+        Returns:
+            True if the parameters are valid, False otherwise.
+        """
```
diegoferigo, line 62:
```suggestion
        """
        pass
```

src/jaxsim/rbda/contacts/common.py
```diff
@@ -0,0 +1,95 @@
...
+
+        Args:
+            position: The position of the collidable point w.r.t. the world frame.
+            velocity:
+                The linear velocity of the collidable point (linear component of the mixed 6D velocity).
+
+        Returns:
+            A tuple containing as first element the computed 6D contact force applied to the contact point and expressed in the world frame,
+            and as second element a tuple of optional additional information.
+        """
```
diegoferigo, line 95:
```suggestion
        """
        pass
```

src/jaxsim/rbda/contacts/common.py
```diff
@@ -0,0 +1,95 @@
...
+
+        Returns:
+            The zero contact state.
+        """
+
+    @abc.abstractmethod
+    def valid(self, **kwargs) -> bool:
+        """
+        Check if the contacts state is valid.
+        """
```
diegoferigo, line 39:
```suggestion
        """
        pass
```

src/jaxsim/rbda/contacts/common.py
```diff
@@ -0,0 +1,95 @@
...
+
+    @classmethod
+    @abc.abstractmethod
+    def zero(cls, **kwargs) -> ContactsState:
+        """
+        Build a zero contact state.
+
+        Returns:
+            The zero contact state.
+        """
```
diegoferigo, line 33:
```suggestion
        """
        pass
```

src/jaxsim/rbda/contacts/common.py
```diff
@@ -0,0 +1,95 @@
...
+
+    @classmethod
+    @abc.abstractmethod
+    def build(cls, **kwargs) -> ContactsState:
+        """
+        Build the contact state object.
+
+        Returns:
+            The contact state object.
+        """
```
diegoferigo, line 23:
```suggestion
        """
        pass
```

diegoferigo (changes requested):
Final suggestions that I overlooked in my previous review:

- Can you double check in our codebase where `JaxSimModel.build` is called? For sure, the new `JaxSimModel.contact_model` attribute has to be copied when the model is reduced [here](https://github.com/ami-iit/jaxsim/blob/35f25b1d006a604257684bc9b89c622d8400ccec/src/jaxsim/api/model.py#L348-L353).
- Instead of `pass`, you can also use `...` but I would not leave the body empty.

src/jaxsim/rbda/contacts/common.py
flferretti: It was to avoid [`unnecessary-pass / W0107`](https://pylint.readthedocs.io/en/latest/user_guide/messages/warning/unnecessary-pass.html)

diegoferigo: Mmh personal taste, I don't like it. That works only because there is a docstring. If you don't like the pass, use `...`. This code is never executed anyway, there's no performance to be optimized, and I like explicitly seeing a no-op body function.

flferretti:
> Can you double check in our codebase where `JaxSimModel.build` is called? For sure, the new `JaxSimModel.contact_model` attribute has to be copied when the model is reduced [here](https://github.com/ami-iit/jaxsim/blob/35f25b1d006a604257684bc9b89c622d8400ccec/src/jaxsim/api/model.py#L348-L353).

The only part in which the `contact_model` is not specified is in the test, yet this is not a problem since the soft contact model is the default value

flferretti:
I added the `pass` statements and the rest of the suggestions in https://github.com/ami-iit/jaxsim/pull/178/commits/ab7a688aba91545e6372f9dac68436dd4efb5b27

diegoferigo:
> > Can you double check in our codebase where `JaxSimModel.build` is called? For sure, the new `JaxSimModel.contact_model` attribute has to be copied when the model is reduced [here](https://github.com/ami-iit/jaxsim/blob/35f25b1d006a604257684bc9b89c622d8400ccec/src/jaxsim/api/model.py#L348-L353).
> 
> The only part in which the `contact_model` is not specified is in the test, yet this is not a problem since the soft contact model is the default value

Ok. Instead, regarding the model reduction, I meant that something to prevent a problem similar to #168 is still missing, right?

flferretti:
> > > Can you double check in our codebase where `JaxSimModel.build` is called? For sure, the new `JaxSimModel.contact_model` attribute has to be copied when the model is reduced [here](https://github.com/ami-iit/jaxsim/blob/35f25b1d006a604257684bc9b89c622d8400ccec/src/jaxsim/api/model.py#L348-L353).
> > 
> > 
> > The only part in which the `contact_model` is not specified is in the test, yet this is not a problem since the soft contact model is the default value
> 
> Ok. Instead, regarding the model reduction, I meant that something to prevent a problem similar to #168 is still missing, right?

It's indeed there https://github.com/ami-iit/jaxsim/blob/dfebf773d924ad6ab67b91891b28cc4281dbc369/src/jaxsim/api/model.py#L369

src/jaxsim/api/model.py
```diff
@@ -50,6 +50,10 @@ class JaxSimModel(JaxsimDataclass):
     def description(self) -> jaxsim.parsers.descriptions.ModelDescription:
         return self._description.get()
 
+    contact_model: jaxsim.rbda.ContactModel | None = dataclasses.field(
+        default=None, repr=False, compare=False, hash=False
+    )
+
```
diegoferigo, line 56: Any reason to add this attribute after the property? Can you please move it after `terrain`?

diegoferigo:
> > > > Can you double check in our codebase where `JaxSimModel.build` is called? For sure, the new `JaxSimModel.contact_model` attribute has to be copied when the model is reduced [here](https://github.com/ami-iit/jaxsim/blob/35f25b1d006a604257684bc9b89c622d8400ccec/src/jaxsim/api/model.py#L348-L353).
> > > 
> > > 
> > > The only part in which the `contact_model` is not specified is in the test, yet this is not a problem since the soft contact model is the default value
> > 
> > 
> > Ok. Instead, regarding the model reduction, I meant that something to prevent a problem similar to #168 is still missing, right?
> 
> It's indeed there
> 
> https://github.com/ami-iit/jaxsim/blob/dfebf773d924ad6ab67b91891b28cc4281dbc369/src/jaxsim/api/model.py#L369

Woops, sorry my bad. All good then, thanks!

diegoferigo (approved):

src/jaxsim/api/model.py
diegoferigo: Also, compare and hash are disabled class-wise, it's no longer necessary to disable them in the field definition.

flferretti: I might have messed something while rebasing. Fixed in https://github.com/ami-iit/jaxsim/pull/178/commits/882a4a0ae04e8028c3ff6b0ab27dbaea43779826

PR #173, opened 2024-06-11
Minor changes for the upcoming release

diegoferigo:
Minor changes that were sitting in my local git repository.

@flferretti feel free to push similar changes, if you have any.

----
📚 Documentation preview 📚: https://jaxsim--173.org.readthedocs.build//173/

PR #179, opened 2024-06-14
Speed up passing different `JaxSimModel` with same pytree structure to JIT-compiled functions

diegoferigo:
PR #173 changed how `JaxSimModel` objects are hashed and compared. It was a necessary change since our frame-related data were read from `JaxSimModel.description` that is a `Static` attribute. Before that PR, the hash of the description was explicitly ignored, but that caused problems when different models were passed to the same JIT-compiled function because JAX was not taking into account possible differences in frame data. The outcome was that the frames test was not passing, and the solution was to make `JaxSimModel.description` properly hashable.

These days I played a bit around with different `JaxSimModels` (e.g. parameterized models), and I discovered that if there are two exact instances `model1` and `model2` created from the same URDF (therefore, not having the second model directly copied from the first one), calling a JIT-compiled function on `model` was extremely slow even if JIT-recompilation was not triggered. **The problem is that in these cases, JAX computes the hash of static attributes**, and right now the hash of `ModelDescription` takes hundreds of milliseconds. This is a problem because the processing of static attributes is orders of magnitude longer than the actual compiled computation.

Originally, I though ways to speed up the equality computation of `ModelDescription`, and this PR contains new `__eq__` methods that do not call `__hash__`.

However, then I realized that the best solution is to move also frame-related data in `KinDynParameters`, similarly to what we already do for links, joints, and contacts. In this way, `JaxSimModel.description` can be ignored again, and there's no longer the need to compute its hash. This speeds up significantly calling function that have been JIT-compiled on `model1` using `model2`.

There still is an overhead. I guess that JAX saves the id of `model1` and skips the check on static attributes, check that instead is done for `model2. The following is the runtime on CPU using two full ErgoCub models:

```python
import jaxsim.api as js
from jaxsim import VelRepr

)

model1 = js.model.JaxSimModel.build_from_model_description(
    model_description=urdf_path,
    is_urdf=True,
)

model2 = js.model.JaxSimModel.build_from_model_description(
    model_description=urdf_path,
    is_urdf=True,
)

data = js.data.random_model_data(model=model1, velocity_representation=VelRepr.Mixed)

# First run for JIT compilation, second one for runtime.
%time _ = js.model.forward_dynamics_aba(model1, data)  # Wall time: 4.31 s
%timeit js.model.forward_dynamics_aba(model1, data)    # 296 µs ± 12.6 µs

# This should not get JIT compiled.
%time _ = js.model.forward_dynamics_aba(model2, data)  # Wall time: 1.56 ms

# This should go almost as fast as on model1.
# The overhead is due to the comparison of static attributes.
%timeit js.model.forward_dynamics_aba(model2, data)  # 883 µs ± 14.2 µs
```

---

Note: this PR goes also in the direction of exploting the on-disk JAX compilation cache supported by the GPU and TPU backends. In the past, I've never managed to make it work with JaxSim, probably due to factors related to this PR. Now things should be better since hash and equality are correct. What I suspect is still missing is wrapping static strings with a custom hash function since by default the hash of a string in Python is not the same among executions.  

----
📚 Documentation preview 📚: https://jaxsim--179.org.readthedocs.build//179/

### DIFF
--- a/src/jaxsim/parsers/descriptions/link.py
+++ b/src/jaxsim/parsers/descriptions/link.py
@@ -12,7 +12,7 @@ import jaxsim.typing as jtp
 from jaxsim.utils import JaxsimDataclass
 
 
-@jax_dataclasses.pytree_dataclass
+@jax_dataclasses.pytree_dataclass(eq=False, unsafe_hash=False)
 class LinkDescription(JaxsimDataclass):
     """
     In-memory description of a robot link.
@@ -40,13 +40,15 @@ class LinkDescription(JaxsimDataclass):
 
     def __hash__(self) -> int:
 
+        from jaxsim.utils.wrappers import HashedNumpyArray
+
         return hash(
             (
                 hash(self.name),
                 hash(float(self.mass)),
-                hash(tuple(np.atleast_1d(self.inertia).flatten().tolist())),
-                hash(int(self.index)) if self.index is not None else 0,
-                hash(tuple(np.atleast_1d(self.pose).flatten().tolist())),
+                HashedNumpyArray.hash_of_array(self.inertia),
+                hash(int(self.index)) if self.index is not None else self.index,
+                HashedNumpyArray.hash_of_array(self.pose),
                 hash(tuple(self.children)),
                 # Here only using the name to prevent circular recursion:
                 hash(self.parent.name) if self.parent is not None else 0,
--- a/src/jaxsim/parsers/descriptions/link.py
+++ b/src/jaxsim/parsers/descriptions/link.py
@@ -47,7 +47,7 @@ class LinkDescription(JaxsimDataclass):
                 hash(self.name),
                 hash(float(self.mass)),
                 HashedNumpyArray.hash_of_array(self.inertia),
-                hash(int(self.index)) if self.index is not None else self.index,
+                hash(int(self.index)) if self.index is not None else 0,
                 HashedNumpyArray.hash_of_array(self.pose),
                 hash(tuple(self.children)),
                 # Here only using the name to prevent circular recursion:
@@ -60,29 +60,19 @@ class LinkDescription(JaxsimDataclass):
         if not isinstance(other, LinkDescription):
             return False
 
-        if self.name != other.name:
-            return False
-
-        if not np.allclose(self.mass, other.mass):
-            return False
-
-        if not np.allclose(self.inertia, other.inertia):
-            return False
-
-        if self.index != other.index:
-            return False
-
-        if not np.allclose(self.pose, other.pose):
-            return False
-
-        if self.children != other.children:
-            return False
-
-        # Here only using the name to prevent circular recursion
-        if self.parent is not None and self.parent.name != other.parent.name:
-            return False
-
-        if self.parent is None and other.parent is not None:
+        if not (
+            self.name == other.name
+            and np.allclose(self.mass, other.mass)
+            and np.allclose(self.inertia, other.inertia)
+            and self.index == other.index
+            and np.allclose(self.pose, other.pose)
+            and self.children == other.children
+            and (
+                (self.parent is not None and self.parent.name == other.parent.name)
+                if self.parent is not None
+                else other.parent is None
+            ),
+        ):
             return False
 
         return True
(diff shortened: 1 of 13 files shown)


src/jaxsim/parsers/descriptions/model.py
```diff
@@ -249,7 +249,25 @@ def __eq__(self, other: ModelDescription) -> bool:
...
+        if self.root != other.root:
+            return False
+
+        if self.joints != other.joints:
+            return False
+
+        if self.frames != other.frames:
+            return False
+
+        if self.root_pose != other.root_pose:
```
flferretti, line 267: We could also make this more compact
```suggestion
        if not all(
            [
                self.name == other.name,
                self.fixed_base == other.fixed_base,
                self.root == other.root,
                self.joints == other.joints,
                self.frames == other.frames,
                self.root_pose == other.root_pose,
            ]
        ):
```

src/jaxsim/parsers/descriptions/link.py
```diff
@@ -58,7 +60,32 @@ def __eq__(self, other: LinkDescription) -> bool:
...
+
+        if self.children != other.children:
+            return False
+
+        # Here only using the name to prevent circular recursion
+        if self.parent is not None and self.parent.name != other.parent.name:
+            return False
+
+        if self.parent is None and other.parent is not None:
+            return False
```
flferretti, line 86: Same as above for compactness
```suggestion
        if not all(
            [
                self.name == other.name,
                np.allclose(self.mass, other.mass),
                np.allclose(self.inertia, other.inertia),
                self.index == other.index,
                np.allclose(self.pose, other.pose),
                self.children == other.children,
                (
                    (self.parent is not None and self.parent.name == other.parent.name)
                    if self.parent is not None
                    else other.parent is None
                ),
            ]
        ):
            return False
```

src/jaxsim/parsers/descriptions/link.py
```diff
@@ -40,13 +40,15 @@ class LinkDescription(JaxsimDataclass):
...
+
         return hash(
             (
                 hash(self.name),
                 hash(float(self.mass)),
-                hash(tuple(np.atleast_1d(self.inertia).flatten().tolist())),
-                hash(int(self.index)) if self.index is not None else 0,
-                hash(tuple(np.atleast_1d(self.pose).flatten().tolist())),
+                HashedNumpyArray.hash_of_array(self.inertia),
+                hash(int(self.index)) if self.index is not None else self.index,
```
flferretti, line 50: Can we rely on `hash(None)`? This should be reproducible only for Python>=3.12 (see https://github.com/python/cpython/pull/99541)

src/jaxsim/utils/wrappers.py
```diff
@@ -76,3 +109,37 @@ def __eq__(self, other: HashedNumpyArray) -> bool:
...
+    @staticmethod
+    def hash_of_array(
+        array: jax.Array | npt.NDArray, precision: float | None = 1e-9
+    ) -> int:
+        """
+        Calculate the hash of a NumPy array.
+
+        Args:
+            array: The array to hash.
+            precision: Optionally limit the precision over which the hash is computed.
```
flferretti, line 122: I'm not that familiar with this logic. Why is this necessary?

flferretti (approved):
LGTM, thanks Diego!

diegoferigo:
Thanks @flferretti for the suggestion on more compact syntax. The original idea was to prevent evaluating all the conditions if one of the first ones is false. Your suggestion using `all` would work similarly only if its argument is a generator expression, which is not the case. I updated your suggestion to use chained `and`, Python stops early also in this case.

src/jaxsim/parsers/descriptions/link.py
diegoferigo: I'm of Python 3.12 and I didn't think about it, if None we can set a 0 hash.

src/jaxsim/utils/wrappers.py
diegoferigo: The logic before this PR had no precision. For low-level reasons I didn't investigate, hashing numpy arrays was only persistent for a couple of seconds, then it was changing to another value. This behaviour puzzled me quite a lot.

The new logic is done in such a way that the hash of numpy arrays is always stable up to the given precision (both for the integer and the decimal parts). And, furthermore, this default should be also compatible in 32-bit precision without overflow.

diegoferigo: In general, hashing numpy arrays is always a bad idea. And this is the reason why `np.all_close` is used for arrays of floating point numbers, that under the hood uses absolute/relative tolerances.

flferretti (approved):

PR #181, opened 2024-06-14
Raise exceptions from jit-compiled functions

diegoferigo:
This PR:

- Introduces a new helper that allows to raise exceptions from within jit-compiled functions.
- For each exception, a new dummy branch triggered by the same condition of the expectin is injected in the compiled code. In this way, the host callback (that can be slow even if no-op) is not triggered unless needed.
- The -apparently- double check on `condition` (both in the callback and the low-level `jax.lax.cond`) is necessary because JAX compiles both branches and it would raise an exception while tracing.

The caveat is that JAX raises a `XlaRuntimeError` to stop the execution of the jit-compiled function. The real exception raised in the callback is printed together with the corresponding stack trace earlier in the output.

Although this method is not capable of handling raised exceptions with a `try` statement (I don't see any way to do that, regardless), at least we can stop the execution by raising.

----
📚 Documentation preview 📚: https://jaxsim--181.org.readthedocs.build//181/

PR #182, opened 2024-06-18
Validate link, frame, and joint indices in our jit-compiled APIs

diegoferigo:
Implements the jit-compatible exceptions of #181 to stop the execution of jit-compiled methods when the integer indices of links, frames, and joints are not valid.

----
📚 Documentation preview 📚: https://jaxsim--182.org.readthedocs.build//182/

### DIFF
--- a/src/jaxsim/api/frame.py
+++ b/src/jaxsim/api/frame.py
@@ -9,6 +9,7 @@ import numpy as np
 import jaxsim.api as js
 import jaxsim.math
 import jaxsim.typing as jtp
+from jaxsim import exceptions
 
 from .common import VelRepr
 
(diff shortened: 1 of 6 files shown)


src/jaxsim/api/frame.py
```diff
@@ -49,19 +60,18 @@ def name_to_idx(model: js.model.JaxSimModel, *, frame_name: str) -> jtp.Int:
...
-    return jnp.array(-1).astype(int)
+    return (
+        jnp.array(
+            np.argwhere(
+                np.array(model.kin_dyn_parameters.frame_parameters.name) == frame_name
+            )
+        )
+        .astype(int)
+        .squeeze()
+    ) + model.number_of_links()
```
flferretti, line 74: We can use the list functions directly. Be aware that `numpy` is no more used
```suggestion
    return (
        jnp.array(
            model.kin_dyn_parameters.frame_parameters.name.index(frame_name)
        ).squeeze()
    ) + model.number_of_links()
```

src/jaxsim/api/link.py
```diff
@@ -31,15 +32,16 @@ def name_to_idx(model: js.model.JaxSimModel, *, link_name: str) -> jtp.Int:
...
+        raise ValueError(f"Link '{link_name}' not found in the model.")
+
+    return (
+        jnp.array(
+            np.argwhere(np.array(model.kin_dyn_parameters.link_names) == link_name)
         )
-    return jnp.array(-1).astype(int)
+        .astype(int)
+        .squeeze()
+    )
```
flferretti, line 44: Same as above. Be aware that `numpy` is no more used
```suggestion
    return jnp.array(model.kin_dyn_parameters.link_names.index(link_name)).squeeze()
```

flferretti (approved):
That's great, thanks! Does it have any concrete effect on the performances?

src/jaxsim/api/frame.py
diegoferigo: I'm not sure if we can index attributes (neither if static) within a jit-compiled function. I can give it a try.

src/jaxsim/api/link.py
diegoferigo: Done in https://github.com/ami-iit/jaxsim/pull/182/commits/039f97fd8ca4095badc1bf57598d925d4c937043

diegoferigo:
> That's great, thanks! Does it have any concrete effect on the performances?

Not really, the modified methods are not called a lot. I don't expect any runtime overhead since the extra branch injected for raising the exception is never called, and the only overhead is a boolean comparison that has almost no cost.

Inspecting the time taken by CI, there's no significant change. It takes 15-20 seconds more on average but this is also because this PR introduces a new test.

PR #185, opened 2024-06-19
Extend test of link bias acceleration

diegoferigo:
I just wanted to make sure that the calculation needed by @xela-95 were correct. Apparently they are, and I took advantage of this check to extend the test.

----
📚 Documentation preview 📚: https://jaxsim--185.org.readthedocs.build//185/

### DIFF
--- a/tests/test_api_link.py
+++ b/tests/test_api_link.py
@@ -223,6 +223,74 @@ def test_link_bias_acceleration(
         Jν_js = js.link.bias_acceleration(model=model, data=data, link_index=index)
         assert pytest.approx(Jν_idt) == Jν_js
 
+    # Test that the conversion of the link bias acceleration works as expected.
+    match data.velocity_representation:
+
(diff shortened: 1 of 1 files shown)


xela-95:
> I just wanted to make sure that the calculation needed by @xela-95 were correct. Apparently they are, and I took advantage of this check to extend the test.

Thanks!!

flferretti (commented):
If there's no rush in merging this, can we think about adding also the test for the mixed representation? Otherwise, I believe it would be nice to refactor the if-condition using `match-case` eventually adding a: 

```python
match VelRepr.Mixed:
    # TODO: [Insert comment here on why we don't test Mixed]
    pass
```

so that we can work on that in the future. What do you think?

diegoferigo:
> can we think about adding also the test for the mixed representation

I don't want to dedicate additional effort towards this now. Merging without mixed.

PR #186, opened 2024-06-21
Finalize minor changes for v0.4 release

flferretti:
This pull request adds minor fixes and updates in preparation for the new release.

@diegoferigo @xela-95 feel free to push similar changes.

----
📚 Documentation preview 📚: https://jaxsim--186.org.readthedocs.build//186/


diegoferigo:
Thanks @flferretti for these minor changes.

@traversaro @flferretti  I'd like to start discussing again about the lock file of pixi. I'm not happy with its size and the size of each modification that occur when it's re-rendered. As far as I understood, there is no real workaround. My proposal is the following:

1. Remove the lock file from the repo.
2. Include the rendered lock file as release artifact, possibly automatically generated from CI.
3. Add in CI a new test that execute the examples from the resulting environment.

The two benefits we get from this approach is that:

1. We avoid large diffs in our git repo and a constantly growing size with little benefit.
2. We don't have to re-render manually the lock file once in a while.
3. Instead of ensuring that the examples from the latest commit can run on a known set of dependencies, we ensure a similar but more loose property on just the releases.

In general, if users clone the repository directly, the intended usage is creating the environment using the `environment.yml`, not `pixi.lock`. Any comment?

traversaro:
Ideally the nice thing of having the `pixi.lock` in the repo is to ensure all the users and CI have the same version when developing (but this assumes that they all use pixi, at least until https://github.com/prefix-dev/pixi/pull/1427 is complete), but clearly these needs to be considered as tradeoff with the other downsides like size. We can discuss this in person next week also with @flferretti ? I wonder if it could make sense to just register `pixi.lock` as a git lfs file? While `git lfs` is typically used for non-binary files, I think it may be a good fit for `pixi.lock` files, and there would be no need for users to actually enable `git lfs` unless they want to use pixi.

diegoferigo:
> Ideally the nice thing of having the `pixi.lock` in the repo is to ensure all the users and CI have the same version when developing (but this assumes that they all use pixi, at least until [prefix-dev/pixi#1427](https://github.com/prefix-dev/pixi/pull/1427) is complete), but clearly these needs to be considered as tradeoff with the other downsides like size. We can discuss this in person next week also with @flferretti ? I wonder if it could make sense to just register `pixi.lock` as a git lfs file? While `git lfs` is typically used for non-binary files, I think it may be a good fit for `pixi.lock` files, and there would be no need for users to actually enable `git lfs` unless they want to use pixi.

Sure, we can align f2f to make a decision. I like the `git lfs` idea as it is a opt-in feature for the fraction of users (arbitrarily small or large) that want to use pixi.

diegoferigo:
@flferretti for the time being, can you remove the pixi lock file from this PR (with a force push, not a commit that reverts the modification). We will handle the lock file separately in a new PR after our f2f discussion.

tests/test_automatic_differentiation.py
```diff
@@ -373,14 +373,14 @@ def test_ad_integration(
...
-        τ: jax.typing.ArrayLike,
-        W_f_L: jax.typing.ArrayLike,
+        W_p_B: jtp.ArrayLike,
+        W_Q_B: jtp.ArrayLike,
+        s: jtp.ArrayLike,
+        W_v_WB: jtp.ArrayLike,
+        ṡ: jtp.ArrayLike,
+        m: jtp.ArrayLike,
+        τ: jtp.ArrayLike,
+        W_f_L: jtp.ArrayLike,
```
diegoferigo, line 383:
```suggestion
        W_p_B: jtp.Vector,
        W_Q_B: jtp.Vector,
        s: jtp.Vector,
        W_v_WB: jtp.Vector,
        ṡ: jtp.Vector,
        m: jtp.Matrix,
        τ: jtp.Vector,
        W_f_L: jtp.Matrix,
```

src/jaxsim/api/data.py
```diff
@@ -625,9 +624,8 @@ def reset_base_pose(self, base_pose: jtp.MatrixLike) -> Self:
 
         W_p_B = base_pose[0:3, 3]
 
-        to_wxyz = np.array([3, 0, 1, 2])
-        W_R_B: jaxlie.SO3 = jaxlie.SO3.from_matrix(base_pose[0:3, 0:3])  # noqa
-        W_Q_B = W_R_B.as_quaternion_xyzw()[to_wxyz]
+        W_R_B: jaxlie.SO3 = jaxlie.SO3.from_matrix(base_pose[0:3, 0:3])
+        W_Q_B = W_R_B.wxyz
```
diegoferigo, line 628: Can we use `jaxsim.math.Quaternion` here?

diegoferigo (changes requested):
Early review on this draft PR. Some comment on patterns used until now in JaxSim:

- About typing, we use `jaxsim.typing.*` to specialize `jax.typing.ArrayLike`. I don't like importing the latter from the former package because it's less clear from where it comes from originally. When `jax.typing.ArrayLike` was used, was because I was lazy to specialize it to e.g. `jtp.Float|jtp.Vector|jtp.Matrix`. Furthermore, the `Like` include plain python types and some more. If we are sure that it's going to be a jax numpy array, let's use the non-`Like` versions. Note that it's long time I'm also planning to remove the `*Jax` types in favour of the non-`Like` ones.
- If not necessary because it saves multiple conversions to `jaxlie` types, let's try to use `jaxsim.math.*` resources in our codebase.

I'll add some commit as soon as you are done. Thanks!

src/jaxsim/api/data.py
flferretti: Do you mean something like `W_Q_B = jaxsim.math.Quaternion.from_dcm(dcm=base_pose[:3][:3])`?

diegoferigo: Yep, when possible I'd like to remove direct usage of `jaxlie` in favor of `jaxsim.math.*` resources. 

```suggestion
        W_Q_B = jaxsim.math.Quaternion.from_dcm(dcm=base_pose[0:3, 0:3])
```

src/jaxsim/api/model.py
```diff
@@ -1361,12 +1363,14 @@ def total_momentum_jacobian(
...
-            B_X_W = jaxlie.SE3.from_matrix(data.base_transform()).inverse().adjoint()
+            B_X_W = Adjoint.from_transform(
+                transform=Transform.inverse(data.base_transform())
+            )
             B_Jh = B_Jh_B @ jax.scipy.linalg.block_diag(B_X_W, jnp.eye(model.dofs()))
 
         case VelRepr.Mixed:
             BW_H_B = data.base_transform().at[0:3, 3].set(jnp.zeros(3))
-            B_X_BW = jaxlie.SE3.from_matrix(BW_H_B).inverse().adjoint()
+            B_X_BW = Adjoint.from_transform(transform=Transform.inverse(BW_H_B))
```
diegoferigo, line 1373:
```suggestion
            B_X_BW = Adjoint.from_transform(transform=BW_H_B, inverse=True)
```

src/jaxsim/parsers/descriptions/link.py
```diff
@@ -106,8 +106,8 @@ def lump_with(
         I_removed = link.inertia
 
         # Create the SE3 object. Note the inverse.
-        r_H_l = jaxlie.SE3.from_matrix(lumped_H_removed).inverse()
-        r_X_l = r_H_l.adjoint()
+        r_H_l = Transform.inverse(lumped_H_removed)
+        r_X_l = Adjoint.from_transform(transform=r_H_l)
```
diegoferigo, line 110:
```suggestion
        r_X_l = Adjoint.from_transform(transform=lumped_H_removed, inverse=True)
```

src/jaxsim/parsers/rod/utils.py
```diff
@@ -50,8 +49,8 @@ def from_sdf_inertial(inertial: rod.Inertial) -> jtp.Matrix:
     L_H_CoM = inertial.pose.transform() if inertial.pose is not None else np.eye(4)
 
     # We need its inverse
-    CoM_H_L = jaxlie.SE3.from_matrix(matrix=L_H_CoM).inverse()
-    CoM_X_L = CoM_H_L.adjoint()
+    CoM_H_L = Transform.inverse(L_H_CoM)
+    CoM_X_L = Adjoint.from_transform(transform=CoM_H_L)
```
diegoferigo, line 53:
```suggestion
    CoM_X_L = Adjoint.from_transform(transform=L_H_CoM, inverse=True)
```

src/jaxsim/math/__init__.py
```diff
@@ -1,11 +1,11 @@
...
 
-from .adjoint import Adjoint
+from .adjoint import Adjoint  # isort:skip
+from .transform import Transform  # isort:skip
 from .cross import Cross
 from .inertia import Inertia
 from .joint_model import JointModel, supported_joint_motion
 from .quaternion import Quaternion
 from .rotation import Rotation
 from .skew import Skew
```
diegoferigo, line 11: Probably it's JointModel that has to be moved as last?

diegoferigo (approved):
Looks good, last minor comments. Let me know when you're done and I might add last minor changes.

diegoferigo:
@flferretti I pushed few changes, let me know if you have any comment. Otherwise this PR on my side is ready.

flferretti:
Thanks @diegoferigo, I'll wait for you to add the fixes for the CI and then I'll change the target branch of #187 to `main`

src/jaxsim/math/transform.py
```diff
@@ -46,8 +46,8 @@ def from_quaternion_and_translation(
 
     @staticmethod
     def from_rotation_and_translation(
-        rotation: jtp.MatrixLike,
-        translation: jtp.VectorLike,
+        rotation: jtp.MatrixLike = jnp.eye(3),
+        translation: jtp.VectorLike = jnp.zeros(3),
```
flferretti, line 50: This may lead to a silent fail, what if we add:
```python
@classmethod 
def unitary(cls):
    return cls.from_rotation_and_translation(
        rotation=jnp.eye(3),
        translation=jnp.zeros(3),
    )
```

```suggestion
        rotation: jtp.MatrixLike,
        translation: jtp.VectorLike,
```

diegoferigo: The main reason for this is to pass only a rotation or only a translation, as it is already done for [`from_quaternion_and_translation`](https://github.com/ami-iit/jaxsim/blob/7340d43172d5ec528f3f7547d532ffeb9770355d/src/jaxsim/math/transform.py#L11-L17) and in the [`Adjoint`](https://github.com/ami-iit/jaxsim/blob/7340d43172d5ec528f3f7547d532ffeb9770355d/src/jaxsim/math/adjoint.py#L63-L67). Of course, if the user calls something named `from_rotation_and_translation` and does not pass a translation, to me it's implicit that the corresponding zero element should be used. I don't see room of misuse in this specific case.

flferretti: Okay, thanks for the explanation. 

Good to go for me, do you have anything else to add to the PR?

src/jaxsim/api/contact.py
```diff
@@ -137,9 +137,17 @@ def collidable_point_dynamics(
...
+    # Import privately the soft contacts classes.
+    from jaxsim.rbda.contacts.soft import SoftContacts, SoftContactsState
+
     # Build the soft contact model.
     match model.contact_model:
-        case s if isinstance(s, SoftContacts):
+
+        case SoftContacts():
+
+            assert isinstance(model.contact_model, SoftContacts)
```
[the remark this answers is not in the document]
diegoferigo: This is to help resources for type hinting like mypy and IDEs. Otherwise they cannot assume the type of `model.contact_model`.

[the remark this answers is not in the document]
flferretti: Okay so I guess this should be part of #190, right? Anyhow we can merge it like this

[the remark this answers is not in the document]
diegoferigo: I don't have plans to work on mypy anytime soon, the issue is mainly a tracker if anyone wants to give it a shot. Curiosity, does your VSCode autocomplete `model.contact_model.parameters`<kbd>Tab</kbd>`K|D|mu`? PyCharm is able to autocomplete (and remove the warnings) only with the asserts.

This is a consequence of having `model.contact_model` typed with the base class `jaxsim.rbda.ContactModel`.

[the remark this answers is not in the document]
flferretti: VSCode doesn't autocomplete with K,D,mu regardless your change, but iPython does, with or without the assert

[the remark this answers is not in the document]
diegoferigo: Ok good to know about VSCode, I always had the impression that PyCharm uses a more advanced engine. IPython is somewhat different because it has access to the runtime variables that can be easily introspected. The IDEs can only access to the textual information, i.e. no runtime information is used. This is because in some occasion you see in Python code either asserts like the one discussed here ([type narrowing](https://mypy.readthedocs.io/en/stable/type_narrowing.html)) or explicit statements like `var: VarType` (that cannot be applied to non-self attributes like in this case).

PR #188, opened 2024-06-21
Speed up computation of contact jacobians

diegoferigo:
The computation of the Jacobians corresponding to the implicit frames associated to the collidable points was introduced in #163. At that time, there was a difference in the body frame considered in the Jacobians computed with `jaxsim.api.model.generalized_free_floating_jacobian`, that was fixed in #167.

For this reason, originally the code to compute the contact Jacobian was calling `jaxsim.link.jacobian` inside a `jax.vmap` transformation. This PR modifies this logic to compute the Jacobians of all links only once, and then with `jax.vmap` extract the one corresponding to the link to which the collidable point is attached.

On the ErgoCub model, that has 32 collidable points, the speedup is significant:

```python
# Before
8.46 ms ± 574 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)

# After
262 µs ± 6.87 µs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)
```

----
📚 Documentation preview 📚: https://jaxsim--188.org.readthedocs.build//188/

### DIFF
--- a/src/jaxsim/api/contact.py
+++ b/src/jaxsim/api/contact.py
@@ -351,17 +351,17 @@ def jacobian(
         output_vel_repr if output_vel_repr is not None else data.velocity_representation
     )
 
-    # For each collidable point, get the Jacobians of their parent link.
+    # Compute the Jacobians of all links.
+    W_J_WL = js.model.generalized_free_floating_jacobian(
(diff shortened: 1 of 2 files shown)


flferretti (approved):
Great, thanks Diego! LGTM

PR #197, opened 2024-07-05
Enable `pre-commit.ci` for style checking

### LINKED ISSUE #195: Consider using `pre-commit.ci` for style checking in CI
By using [`pre-commit.ci`](https://pre-commit.ci/), we could eliminate the file related to the `style` CI configuration, e.g. [.github/workflows/style.yml](https://github.com/ami-iit/jaxsim/blob/main/.github/workflows/style.yml) and substitute it with something that leverages the already present `.pre-commit-config.yml` and the configuration from the `pyproject.toml`.

This would allow us to add additional checks (see https://github.com/pre-commit/pre-commit-hooks), especially the `check-added-large-files`, which would prevent the user to commit the `pixi.lock` pulled from `lfs`, and the others checks already present in the pre-commit configuration.

flferretti:
This PR removes the current style checking from GitHub Actions and moves it to `pre-commit.ci`, this will allow for additional checking.

Closes #195 

----
📚 Documentation preview 📚: https://jaxsim--197.org.readthedocs.build//197/


diegoferigo:
Before reviewing this PR, a curiosity. Is there any way to perform this check without relying on external services? And I mean, can we do it using a regular GitHub workflow?

flferretti:
> Before reviewing this PR, a curiosity. Is there any way to perform this check without relying on external services? And I mean, can we do it using a regular GitHub workflow?

Yes, there is the lite version of the CI, you can find more details at https://pre-commit.ci/lite.html

flferretti:
Btw I temporarily disabled the pre-commit.ci integration from the repo as it got super annoying with the automatic commit fixes

diegoferigo:
If there is an option providing all the features we need that doesn't use any external service, I'd rather go towards that way. Many years ago, we were interfacing with similar external services (like CirleCI, TravisCI, etc), and they are definitely more prone to get broken, be down, or have some sort of problems.

Would it require a big effort enabling the lite version as GitHub Workflow?

flferretti:
> Would it require a big effort enabling the lite version as GitHub Workflow?

Totally not, I'll add that

flferretti:
After discussing with @diegoferigo, we decided to use the `pre-commit.ci` instead of the GitHub Actions integration as it is faster and it is actively maintained.

I pushed a dummy commit to verify that the CI actually fails when the formatting is wrong. This can be fixed in three ways:
1. use `pre-commit` hook (suggested)
2. trigger the bot commenting with `pre-commit.ci autofix`
3. adding a `pre-commit.ci autofix` label (to avoid)

flferretti:
pre-commit.ci autofix

flferretti:
pre-commit.ci autofix

flferretti:
I dropped the last commit, for some reason the bot doesn't work. It may be due to some issue upstream, but I'll verify that. Ready for review @diegoferigo, thanks a lot!

.github/workflows/ci_cd.yml
```diff
@@ -28,7 +28,7 @@ jobs:
       - name: Set up Python
         uses: actions/setup-python@v5
         with:
-          python-version: "3.11"
+          python-version-file: "pyproject.toml"
```
diegoferigo, line 31: The `jobs.package` in theory can run on any Python version (also the latest one) since it does not run any code related to the JaxSim package. The version stored in `pyproject.toml` is instead related to the runtime version required by JaxSim.

I don't see any problem in reading it here (it would work regardless) but it has in the same time no real advantage. This job should be successful also using Python versions older than `pyproject.toml@project.requires-python`.

---

- https://github.com/actions/setup-python/blob/main/docs/advanced-usage.md#using-the-python-version-file-input
- https://github.com/ami-iit/jaxsim/actions/runs/9858438537/job/27219807789?pr=197#step:3:9

.pre-commit-config.yaml
```diff
@@ -1,3 +1,11 @@
+ci:
+  autofix_prs: false
+  autoupdate_schedule: quarterly
+  submodules: false
+
+default_language_version:
+  python: python3.11
```
diegoferigo, line 7: Since we support python 3.10 as minimum version, should we use 3.10 also here?

diegoferigo (approved):

diegoferigo:
Can you please also cherry pick the updates of #200 here?

.pre-commit-config.yaml
flferretti: I discovered that for some reason, `pre-commit` is not able to build a venv from a specific python version, therefore I just specified `default_language_version: python:python3`

.github/workflows/ci_cd.yml
flferretti: Alright! Do you want me to drop that commit?

diegoferigo: I don't have a strong preference, what do you think? Probably having it explicit as before may be more clear as it does not raise the doubt of which version gets installed?

.pre-commit-config.yaml
diegoferigo: I have never played that much with pre-commit, a question. Is the `venv` created automatically? Shouldn't it be specified as hook in the repos that are listed in the yaml file?

diegoferigo: Ok now I understood, I think. 

- pre-commit always create a virtual environment.
- virtualenvs can be created only with a matching version of the corresponding `python` executable, that is generally the system's python (but it could be the conda's python if the environment is enabled).
- you need to pass a generic `python3` because you cannot make assumptions about what python version is currently active.

https://pre-commit.com/#overriding-language-version
-

.github/workflows/ci_cd.yml
flferretti: Yes it makes sense!

.pre-commit-config.yaml
flferretti: Okay, got it. Thanks for the explanation!

.github/workflows/ci_cd.yml
flferretti: Done with rebase https://github.com/ami-iit/jaxsim/compare/c6b9cea8232746180b43fb99acf702a3baf4d995..b79406c2f8805a8585039387d07e754e1333bafa

flferretti:
> Can you please also cherry pick the updates of #200 here?

Done in https://github.com/ami-iit/jaxsim/pull/197/commits/b79406c2f8805a8585039387d07e754e1333bafa.

Would you like to add anything? Otherwise I'll merge this

diegoferigo (approved):
Nope, good to go, thanks!

flferretti:
FYI @Giulero

PR #201, opened 2024-07-09
Add `W_X_i` and `W_v_Wi` function parameters in `process_point_kinematics`

xela-95:
I've noted that in this function of `collidable_points.py` the two variables were used directly from the outer scope, with potential undefined behavior.

----
📚 Documentation preview 📚: https://jaxsim--201.org.readthedocs.build//201/

### DIFF
--- a/src/jaxsim/rbda/collidable_points.py
+++ b/src/jaxsim/rbda/collidable_points.py
@@ -118,7 +118,7 @@ def collidable_points_pos_vel(
     # ==================================================
 
     def process_point_kinematics(
-        Li_p_C: jtp.Vector, parent_body: jtp.Int
+        Li_p_C: jtp.Vector, parent_body: jtp.Int, W_X_i: jtp.Matrix, W_v_Wi: jtp.Matrix
     ) -> tuple[jtp.Vector, jtp.Vector]:
 
         # Compute the position of the collidable point.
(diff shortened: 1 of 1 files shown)


flferretti (approved):
Thanks a lot @xela-95! LGTM

diegoferigo (approved):
This is a pattern that is often used in lambda functions (and, generally, closures). In this specific case, those variables are already available when the function is created, and the function body is closed over these two variables that are used only as read-only.

I don't see any problem in this case resulting from the spurious access of global variables (as it happened in your scripts recently). Anyway, this change does not affect performance, if it enhances readability, let's proceed merging. Just note that this pattern was used on purpose and it is also used similarly elsewhere.

xela-95:
> This is a pattern that is often used in lambda functions (and, generally, closures). In this specific case, those variables are already available when the function is created, and the function body is closed over these two variables that are used as read-only.
> 
> I don't see any problem in this case resulting from the spurious access of global variables (as it happened in your scripts recently). Anyway, this change does not affect performance, if it enhances readability, let's proceed merging. Just note that this pattern was used on purpose and it is also used similarly elsewhere.

Thanks, in this case I didn't know if it was used on purpose or it was a code smell, and I opened this PR for this reason. I leave to you to decide (or not) to merge it :)

diegoferigo:
I think I'll close this PR for the time being. It's a pattern used thoroughly in JaxSim (often on `model` and `data`, but sometimes on other variables).

I want to remark again that scoping problems may occur only when global variables are involved, and this is pretty much the reason why there is the consensus that global variables should be used the least possible. Of course, in Python scripts, it's often quicker to just develop all global (note that also using `if __name__ == "__main__"` is global), the real solution is to wrap the content in a `def main()` function.

PR #204, opened 2024-07-11
Fix `test_simulations::test_box_with_zero_gravity`

### LINKED ISSUE #203: `test_simulation::test_box_with_zero_gravity` generates NaNs in `VelRepr.Inertial` with 10N force
When defining the applied forces in https://github.com/ami-iit/jaxsim/blob/2056e707805d7281c48b19574bf1d588e3c88e93/tests/test_simulations.py#L118-L122

if I scale up the sampling range from $\verb |[-1, 1]| N$ to $\verb |[-10, 10]| N$, the base position result in an array of NaNs only in inertial velocity representation. The only solution in to avoid NaNs is to reduce the integration step size, yet this should not be necessary.

diegoferigo:
Closes #203.

----
📚 Documentation preview 📚: https://jaxsim--204.org.readthedocs.build//204/

### DIFF
--- a/tests/test_simulations.py
+++ b/tests/test_simulations.py
@@ -6,7 +6,7 @@ import jaxsim.api as js
 import jaxsim.integrators
 import jaxsim.rbda
 from jaxsim import VelRepr
-from jaxsim.rbda.contacts.soft import SoftContactsParams
+from jaxsim.utils import Mutability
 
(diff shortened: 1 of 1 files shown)


flferretti (approved):
Thanks, LGTM!

PR #206, opened 2024-07-12
Improve custom cameras support in `MujocoVisualizer` and `MujocoVideoRecorder`

diegoferigo:
- The `MujocoVisualizer` can now be opened with a configuration[^1] that initializes the camera towards a specific target. Then, the user is free to roam around interactively.
- Users can now create more generic custom cameras from an extended set of parameters[^2].
- The `jaxsim.mujoco.RodModelToMjcf.convert` function now accepts also `MujocoCamera` objects of a sequence of such objects, removing the need to serialize them to dictionary. However, for the most generic camera support, the method still accepts a plain dictionary or list of dictionaries.
- The default "track" camera of the `MujocoVideoRecorder` is not more explicit as it is defined as default argument.

I found interesting the approach used to specify the initial viewpoint of the interactive viewer. It seemed to me much easier to setup than the parameters necessary to create a custom camera. We might think to understand what is the mapping logic between them, and provide a new `MujocoCamera.build_from_target_view` helper.

[^1]: https://mujoco.readthedocs.io/en/stable/XMLreference.html#visual-global
[^2]: https://mujoco.readthedocs.io/en/stable/XMLreference.html#body-camera


src/jaxsim/mujoco/loaders.py
```diff
@@ -598,21 +611,106 @@ def convert(
...
+            mode="fixed",
+            fovy="60",
+            pos=" ".join(p.astype(str).tolist()),
+            quat=" ".join(Q.astype(str).tolist()),
+            **kwargs,
+        )
+
+    def asdict(self) -> dict[str, str]:
+
+        return {k: v for k, v in dataclasses.asdict(self).items() if v is not None}
```
flferretti, line 716: Wouldn't it be a bit confusing to silently exclude the `None` values? Probably this logic can fit better in the `__repr__` instead

src/jaxsim/mujoco/loaders.py
flferretti, line 1: Can you please remove the double `numpy` import in line 263?

src/jaxsim/mujoco/loaders.py
```diff
@@ -598,21 +611,106 @@ def convert(
...
+        W_H_C[0:3, 0:3] = W_H_C[0:3, 0:3] @ R_el
+
+        # Process distance.
+        tf_distance = np.eye(4)
+        tf_distance[2, 3] = distance
+        W_H_C = W_H_C @ tf_distance
+
+        # Extract the position and the quaternion.
+        p = W_H_C[0:3, 3]
+        Q = Rotation.from_matrix(W_H_C[0:3, 0:3]).as_quat()[np.array([3, 0, 1, 2])]
```
flferretti, line 703:
```suggestion
        Q = Rotation.from_matrix(W_H_C[0:3, 0:3]).as_quat(scalar_first=True)
```

src/jaxsim/mujoco/loaders.py
```diff
@@ -598,21 +611,106 @@ def convert(
...
+    fovy: str = "45"
+    pos: str = "0 0 0"
+
+    quat: str | None = None
+    axisangle: str | None = None
+    xyaxes: str | None = None
+    zaxis: str | None = None
+    euler: str | None = None
+
+    name: str | None = None
```
flferretti, line 627: Should we somehow specify that the usage of some attributes excludes the usage of other attributes?

flferretti (commented):
Thanks! I left some comments

src/jaxsim/mujoco/loaders.py
diegoferigo: I wrote this method before adding to [`RodModelToMjcf.convert`](https://github.com/ami-iit/jaxsim/blob/e9a89e32f45aff03bef48e27ac8be3e28b592b71/src/jaxsim/mujoco/loaders.py#L170-L171) the support of receiving `MujocoCamera|Sequence[MujocoCamera]`. Before this update, the method was taking a dictionary that, if generated using `dataclasses.asdict` from a `MujocoCamera` object, included the `None` elements. The conversion to XML of such dictionary was failing, the None attributes had to be ignored.

What do you suggest? I can remove this method now, it's no longer needed.

src/jaxsim/mujoco/loaders.py
diegoferigo: Sure, I didn't notice, probably a leftover from an old PR.

src/jaxsim/mujoco/loaders.py
diegoferigo: Right, the `Rotation` from plain SciPy has this kind-of new option. I also need to add `scipy` as `viz` dependency.

src/jaxsim/mujoco/loaders.py
diegoferigo: I guess that we can redirect users to the Mujoco documentation as already done elsewhere in this PR. I don't want to dedicate any time to copy-paste complex documentation. Does it sound good?

src/jaxsim/mujoco/loaders.py
flferretti: Thanks! Do you think it would make sense to just use `jaxlie` or `jaxsim.math.Rotation` to avoid adding an additional dependency? Otherwise we should also push a new build to conda-forge

src/jaxsim/mujoco/loaders.py
flferretti: Yes, I guess that users can eventually use `dataclasses.asdict` for this

src/jaxsim/mujoco/loaders.py
diegoferigo: Done in https://github.com/ami-iit/jaxsim/pull/206/commits/6b86a1527072578eca4a26863b45526bf7aef228.

src/jaxsim/mujoco/loaders.py
diegoferigo: Done in https://github.com/ami-iit/jaxsim/pull/206/commits/6b86a1527072578eca4a26863b45526bf7aef228.

src/jaxsim/mujoco/loaders.py
diegoferigo: Done in https://github.com/ami-iit/jaxsim/pull/206/commits/6b86a1527072578eca4a26863b45526bf7aef228.

src/jaxsim/mujoco/loaders.py
diegoferigo: For the time being, I'd stick with plain `scipy`. If you notice, the `jaxsim.mujoco` package is standalone, it does not depend on any jax-based library. This was done on purpose if, at some point, we want to port it outside JaxSim. Scipy is not a lightweight dependency, but it will be likely already installed in the majority of user's setups.

> Otherwise we should also push a new build to conda-forge

This needs to be done not before the next tagged release right?

src/jaxsim/mujoco/loaders.py
diegoferigo: > > I can remove this method now, it's no longer needed.
> 
> Yes, I guess that users can eventually use `dataclasses.asdict` for this

Note that if there are `None` attributes, the dictionary produced with `dataclasses.asdict` won't be valid when fed into `lxml` since it would include in the XML attributes with wrong content not supported by the schema.

diegoferigo: Woops forget everything I wrote. I forgot I was using this method [here](https://github.com/ami-iit/jaxsim/blob/f0e6f0c1b075e664cfcbc1236729ba6035594d09/src/jaxsim/mujoco/loaders.py#L499). Brought it back.

src/jaxsim/mujoco/loaders.py
flferretti: > For the time being, I'd stick with plain scipy

Alright!


> This needs to be done not before the next tagged release right?

Yes, we can publish a new build without bumping the version

diegoferigo: > > This needs to be done not before the next tagged release right?
> 
> Yes, we can publish a new build without bumping the version

Yep but this was not required by previous versions if I don't mistake. It's something new here.

diegoferigo:
I spotted a bug in the computation of camera parameters from the desired view. Solved in https://github.com/ami-iit/jaxsim/compare/f0e6f0c1b075e664cfcbc1236729ba6035594d09..b25c33837794a796eda85e1121f6327b379e47c8.

environment.yml
```diff
@@ -5,7 +5,7 @@ dependencies:
   # ===========================
   # Dependencies from setup.cfg
   # ===========================
-  - python=3.11
+  - python >= 3.12.0
```
flferretti, line 8: Why is this necessary?

diegoferigo: This was just a maintenance dependecy bump. Since readthedocs built successfully, I thought why not updating to the latest python version. Did you have any reason to stay one version behind?

flferretti: It's okay to update the default Python version used in the build environment, I was just wondering if it was a constraint for `scipy`

diegoferigo: Nope no reason. I was just passing through the yaml file and I saw this pinning that could have been updated.

src/jaxsim/mujoco/loaders.py
flferretti: Done in https://github.com/conda-forge/jaxsim-feedstock/pull/6

PR #209, opened 2024-07-17
Add the possibility to set forces to frames

### LINKED ISSUE #159: Extend `JaxSimModelReferences` to support applying forces to frames
We already have [`JaxSimModelReferences.apply_link_forces`](https://github.com/ami-iit/jaxsim/blob/7cfca207e688ebffa9690ccdd6f225336207b4d7/src/jaxsim/api/references.py#L316-L324) to apply forces to the links. Now that frame support is finalized (#147), it would be great adding a new `JaxSimModelReferences.apply_frame_force` method that allows to apply a 6D force to a given frame. In this way, the logic can directly detect the parent link of the frame and automatically express the 6D frame force to the frame of the parent link.

flferretti:
The implementation extends `JaxSimModelReferences` with a `apply_frame_forces` method that first converts the forces to inertial representation, then converts them to the parent link forces and finally calls `apply_link_forces` to return the object with the updated link forces. 
The conversion logic can be found at https://github.com/ami-iit/jaxsim/issues/159#issuecomment-2233371413.

Closes #159 

----
📚 Documentation preview 📚: https://jaxsim--209.org.readthedocs.build//209/

### DIFF
--- a/src/jaxsim/api/references.py
+++ b/src/jaxsim/api/references.py
@@ -8,6 +8,7 @@ import jax_dataclasses
 
 import jaxsim.api as js
 import jaxsim.typing as jtp
+from jaxsim import exceptions
 from jaxsim.math import Adjoint
 from jaxsim.utils.tracing import not_tracing
 
@@ -496,9 +497,10 @@ class JaxSimModelReferences(js.common.ModelDataWithVelocityRepresentation):
             model=model, frame_idxs=frame_idxs
         )
 
-        if not_tracing(forces) and not data.valid(model=model):
-            raise ValueError("The provided data is not valid for the model")
-
+        exceptions.raise_value_error_if(
+            condition=not_tracing(forces) and not data.valid(model=model),
+            msg="The provided data is not valid for the model",
+        )
         W_H_Fi = jax.vmap(
             lambda frame_idx: js.frame.transform(
                 model=model, data=data, frame_index=frame_idx
@@ -507,9 +509,7 @@ class JaxSimModelReferences(js.common.ModelDataWithVelocityRepresentation):
 
         # Helper function to convert a single 6D force to the inertial representation
         # considering as body the frame (i.e. L_f_F and LW_f_F).
-        def to_inertial(
-            f_F: jtp.MatrixLike, W_H_F: jtp.MatrixLike, frame_idx: jtp.ArrayLike
-        ) -> jtp.Matrix:
+        def to_inertial(f_F: jtp.MatrixLike, W_H_F: jtp.MatrixLike) -> jtp.Matrix:
             return JaxSimModelReferences.other_representation_to_inertial(
                 array=f_F,
                 other_representation=self.velocity_representation,
@@ -527,17 +527,14 @@ class JaxSimModelReferences(js.common.ModelDataWithVelocityRepresentation):
             case _:
                 raise ValueError("Invalid velocity representation.")
 
+        W_H_L = js.model.forward_kinematics(model=model, data=data)
+
         def convert_to_link_force(
             W_f_F: jtp.MatrixLike, W_H_F: jtp.MatrixLike, parent_link_idx: jtp.ArrayLike
         ) -> jtp.Matrix:
-            W_H_L = js.link.transform(
-                model=model, data=data, link_index=parent_link_idx
-            )
+            L_Xf_W = Adjoint.from_transform(W_H_L[parent_link_idx]).T
 
-            F_H_L = js.transform.inverse(W_H_F) @ W_H_L
-            F_X_L = Adjoint.from_transform(transform=F_H_L, inverse=True)
-
-            return F_X_L @ W_f_F
+            return L_Xf_W @ W_f_F
 
         W_f_L_i = jax.vmap(convert_to_link_force)(W_f_F, W_H_Fi, parent_link_idxs)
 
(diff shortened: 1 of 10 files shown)


src/jaxsim/api/references.py
```diff
@@ -441,3 +442,111 @@ def convert_using_link_frame(
...
+        W_H_Fi = jax.vmap(
+            lambda frame_idx: js.frame.transform(
+                model=model, data=data, frame_index=frame_idx
+            )
+        )(frame_idxs)
+
+        # Helper function to convert a single 6D force to the inertial representation
+        # considering as body the frame (i.e. L_f_F and LW_f_F).
+        def to_inertial(
+            f_F: jtp.MatrixLike, W_H_F: jtp.MatrixLike, frame_idx: jtp.ArrayLike
```
diegoferigo, line 507: It seems to me that the frame index is unused, can we remove it?

src/jaxsim/api/references.py
```diff
@@ -441,3 +442,111 @@ def convert_using_link_frame(
...
+
+        return self.apply_link_forces(
+            model=model,
+            data=data,
+            link_names=js.link.idxs_to_names(
+                model=model, link_indices=parent_link_idxs
+            ),
+            forces=W_f_L,
+            additive=additive,
+        )
```
diegoferigo, line 552: Since you're passing inertial-fixed forces, you need to use `self` within a context manager that switches to `VelRepr.Inertial`.

src/jaxsim/api/references.py
```diff
@@ -441,3 +442,111 @@ def convert_using_link_frame(
...
+            W_f_F: jtp.MatrixLike, W_H_F: jtp.MatrixLike, parent_link_idx: jtp.ArrayLike
+        ) -> jtp.Matrix:
+            W_H_L = js.link.transform(
+                model=model, data=data, link_index=parent_link_idx
+            )
+
+            F_H_L = js.transform.inverse(W_H_F) @ W_H_L
+            F_X_L = Adjoint.from_transform(transform=F_H_L, inverse=True)
+
+            return F_X_L @ W_f_F
```
diegoferigo, line 536: It seems to me that you are transforming forces with a wrong $\mathbf{X}$, can you please double check? Here below my suggestion: 

```suggestion

        W_H_L = js.model.forward_kinematic(model=model, data=data)

        def convert_to_link_force(
            W_f_F: jtp.MatrixLike, W_H_F: jtp.MatrixLike, parent_link_idx: jtp.ArrayLike
        ) -> jtp.Matrix:
   
            W_Xv_L = Adjoint.from_transform(transform=W_H_L[parent_link_idx, :, :])
            L_Xf_W = W_Xv_L.T

            return L_Xf_W @ W_f_F
```

diegoferigo (changes requested):
Thanks @flferretti for taking care of this new feature! I have some doubts on the calculation of the forces, I provided some extra context in https://github.com/ami-iit/jaxsim/issues/159#issuecomment-2244545265.

src/jaxsim/api/references.py
flferretti: You're right! Done in https://github.com/ami-iit/jaxsim/pull/209/commits/cf74cfffcd101e1d08f802c03e14ec6f078e9129

src/jaxsim/api/references.py
flferretti: Totally yes! Having [F841](https://docs.astral.sh/ruff/rules/unused-variable/) disabled [here](https://github.com/ami-iit/jaxsim/blob/4189a2c4e794ec3e191e660cc9f2285522fe9ce1/pyproject.toml#L159) makes this type of error invisible. Addressed in https://github.com/ami-iit/jaxsim/pull/209/commits/cf74cfffcd101e1d08f802c03e14ec6f078e9129

src/jaxsim/api/references.py
```diff
@@ -441,3 +442,111 @@ def convert_using_link_frame(
...
+            raise ValueError(msg.format(len(frame_names), f_F.shape[0]))
+
+        # Extract the frame indices.
+        frame_idxs = js.frame.names_to_idxs(frame_names=frame_names, model=model)
+        parent_link_idxs = js.frame.idx_of_parent_link(
+            model=model, frame_idxs=frame_idxs
+        )
+
+        if not_tracing(forces) and not data.valid(model=model):
+            raise ValueError("The provided data is not valid for the model")
```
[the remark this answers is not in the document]
flferretti: Absolutely! Done in https://github.com/ami-iit/jaxsim/pull/209/commits/cf74cfffcd101e1d08f802c03e14ec6f078e9129

src/jaxsim/api/references.py
flferretti: Yes, I made a mistake in #159. Thanks a lot! Addressed in https://github.com/ami-iit/jaxsim/pull/209/commits/cf74cfffcd101e1d08f802c03e14ec6f078e9129

src/jaxsim/api/references.py
flferretti: I added https://github.com/ami-iit/jaxsim/pull/209/commits/f30d3cf84bade8a4e35671c085609a7736f53cfb to prevent this error in the future. I needed to add a lot of `noqa`, so it kinda makes the code base a bit less clean. Let me know if you want me to drop it

src/jaxsim/api/references.py
```diff
@@ -445,3 +447,110 @@ def convert_using_link_frame(
...
+        # Extract the frame indices.
+        frame_idxs = js.frame.names_to_idxs(frame_names=frame_names, model=model)
+        parent_link_idxs = js.frame.idx_of_parent_link(
+            model=model, frame_idxs=frame_idxs
+        )
+
+        exceptions.raise_value_error_if(
+            condition=not_tracing(forces) and not data.valid(model=model),
+            msg="The provided data is not valid for the model",
+        )
```
diegoferigo, line 503: The `not_tracing` was a super dirty workaround that we no longer need now with real jit-compatible exceptions. We can leave just the check on data validity.

```suggestion
        exceptions.raise_value_error_if(
            condition=jnp.logical_not(data.valid(model=model)),
            msg="The provided data is not valid for the model",
        )

```

src/jaxsim/api/references.py
diegoferigo: Having the `noqa` is not a problem on my side, I prefer keeping the extra assigned variable for readability, it saves (at least to me) mental bandwidth when I review the code -even if it's mine xD-.

src/jaxsim/api/references.py
diegoferigo: There should not be any inverse, the relation between the transform for 6D velocities and 6D forces is the following:

image

src/jaxsim/api/references.py
flferretti: I totally agree with that!

diegoferigo (approved):

src/jaxsim/api/references.py
flferretti: Addressed in https://github.com/ami-iit/jaxsim/pull/209/commits/92cb0bb0b75b1f4dcb983aa46a8172a3fa03c3e3

PR #210, opened 2024-07-17
Add function to compute the Jacobian derivative of additional frames rigidly attached to links

### LINKED ISSUE #208: Add frame Jacobian derivative
For the calculation of the acceleration of a frame, it would be useful if Jaxsim could calculate the Jacobian derivative for frames, as it already does for [links](https://github.com/ami-iit/jaxsim/blob/ba9b8da97f72fd51ddc2a1c771991f89c9a8a9a4/src/jaxsim/api/link.py#L389).

xela-95:
Closes #208 

----
📚 Documentation preview 📚: https://jaxsim--210.org.readthedocs.build//210/

### DIFF
--- a/src/jaxsim/api/frame.py
+++ b/src/jaxsim/api/frame.py
@@ -399,7 +399,7 @@ def jacobian_derivative(
 
         case VelRepr.Mixed:
             W_H_F = transform(model=model, data=data, frame_index=frame_index)
-            W_H_FW = W_H_F.at[0:3, 0:3].set(jnp.zeros((3, 3)))
+            W_H_FW = W_H_F.at[0:3, 0:3].set(jnp.eye(3))
             FW_H_W = Transform.inverse(W_H_FW)
             O_X_W = FW_X_W = Adjoint.from_transform(transform=FW_H_W)
             with data.switch_velocity_representation(VelRepr.Mixed):
--- a/src/jaxsim/api/frame.py
+++ b/src/jaxsim/api/frame.py
@@ -376,7 +376,6 @@ def jacobian_derivative(
         case VelRepr.Inertial:
             O_X_W = W_X_W = Adjoint.from_transform(transform=jnp.eye(4))
             O_Ẋ_W = W_Ẋ_W = jnp.zeros((6, 6))
-            # O_J̇_WF_I = W_J̇_WL_I
             O_J̇_WF_I = jnp.zeros(shape=(6, 6 + model.dofs()))
             O_J̇_WF_I += O_Ẋ_W @ W_J_WL_W @ T
             O_J̇_WF_I += O_X_W @ W_J̇_WL_W @ T
@@ -391,7 +390,6 @@ def jacobian_derivative(
             W_v_WF = W_J_WL_W @ W_nu
             W_vx_WF = Cross.vx(W_v_WF)
             O_Ẋ_W = F_Ẋ_W = -F_X_W @ W_vx_WF
-            # O_J̇_WF_I = F_X_W @ (W_J̇_WL_I - W_vx_WF @ W_J_WL_I)
             O_J̇_WF_I = jnp.zeros(shape=(6, 6 + model.dofs()))
             O_J̇_WF_I += O_Ẋ_W @ W_J_WL_W @ T
             O_J̇_WF_I += O_X_W @ W_J̇_WL_W @ T
@@ -411,10 +409,7 @@ def jacobian_derivative(
                 )
                 FW_v_WF = FW_J_WF_FW @ data.generalized_velocity()
             W_v_W_FW = jnp.zeros(6).at[0:3].set(FW_v_WF[0:3])
-
             W_vx_W_FW = Cross.vx(W_v_W_FW)
-
-            # O_J̇_WF_I = FW_X_W @ (W_J̇_WL_I - W_vx_WFW @ W_J_WL_I)
             O_Ẋ_W = FW_Ẋ_W = -FW_X_W @ W_vx_W_FW
 
             O_J̇_WF_I = jnp.zeros(shape=(6, 6 + model.dofs()))
--- a/src/jaxsim/api/frame.py
+++ b/src/jaxsim/api/frame.py
@@ -335,8 +335,14 @@ def jacobian_derivative(
             W_H_FW = W_H_F.at[0:3, 0:3].set(jnp.zeros((3, 3)))
             FW_H_W = Transform.inverse(W_H_FW)
             FW_X_W = Adjoint.from_transform(transform=FW_H_W)
-            W_v_WF = W_J_WL_I @ data.generalized_velocity()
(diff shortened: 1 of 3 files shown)


xela-95:
In https://github.com/ami-iit/jaxsim/pull/210/commits/b747a26246927128531f7062533d2c809c3aafd0 I commented out the test against iDynTree and tried to test directly $\dot J$ against the same quantity computing through AD. It seems that the inertial and body cases pass for all models, while the mixed case no.

xela-95:
In https://github.com/ami-iit/jaxsim/pull/210/commits/78b690d2bbfd3a046e025bd69dd6e55b52df2ab8 I added also explicitly handled the conversion of the input representation of the jacobian via matrix ${} ^W T _I = diag({} ^W X _I, I _n)$ (with $I$ the active data representation). Unfortunately the test still fails for the `[UR10-mixed]` and `[erogcub-mixed]` cases, while it works for `[box-mixed]`. This is making me think of something related to joints that the box does not have, but I could not spot the bug here.

@diegoferigo @flferretti any idea?

src/jaxsim/api/frame.py
```diff
@@ -257,3 +257,172 @@ def jacobian(
...
+            O_Ẋ_W = F_Ẋ_W = -F_X_W @ W_vx_WF
+            # O_J̇_WF_I = F_X_W @ (W_J̇_WL_I - W_vx_WF @ W_J_WL_I)
+            O_J̇_WF_I = jnp.zeros(shape=(6, 6 + model.dofs()))
+            O_J̇_WF_I += O_Ẋ_W @ W_J_WL_W @ T
+            O_J̇_WF_I += O_X_W @ W_J̇_WL_W @ T
+            O_J̇_WF_I += O_X_W @ W_J_WL_W @ Ṫ
+
+        case VelRepr.Mixed:
+            W_H_F = transform(model=model, data=data, frame_index=frame_index)
+            W_H_FW = W_H_F.at[0:3, 0:3].set(jnp.zeros((3, 3)))
```
diegoferigo, line 402:
```suggestion
            W_H_FW = W_H_F.at[0:3, 0:3].set(jnp.eye(3))
```

diegoferigo (changes requested):
I need some time to go through you code, but the following suggestion is definitely a bug. Let's see what happens as soon as you fix it.

src/jaxsim/api/frame.py
xela-95: It was that indeed!! Thank you so much, I was thinking of rocket science and it was a dummy error instead :)

diegoferigo: Ahah a set of fresh eyes are often useful :eyes: I'm glad it helped, it was really a 10 seconds look :) I'm super happy that the approach with the three-derivatives components worked well also in this case.

src/jaxsim/api/frame.py
```diff
@@ -257,3 +257,167 @@ def jacobian(
...
+            O_Ẋ_W = W_Ẋ_W = jnp.zeros((6, 6))
+            O_J̇_WF_I = jnp.zeros(shape=(6, 6 + model.dofs()))
+            O_J̇_WF_I += O_Ẋ_W @ W_J_WL_W @ T
+            O_J̇_WF_I += O_X_W @ W_J̇_WL_W @ T
+            O_J̇_WF_I += O_X_W @ W_J_WL_W @ Ṫ
+
+        case VelRepr.Body:
+            W_H_F = transform(model=model, data=data, frame_index=frame_index)
+            F_H_W = Transform.inverse(W_H_F)
+            O_X_W = F_X_W = Adjoint.from_transform(transform=F_H_W)
```
diegoferigo, line 387:
```suggestion
            O_X_W = F_X_W = Adjoint.from_transform(transform=W_H_F, inverse=True)
```

src/jaxsim/api/frame.py
```diff
@@ -257,3 +257,167 @@ def jacobian(
...
+                )
+                FW_v_WF = FW_J_WF_FW @ data.generalized_velocity()
+            W_v_W_FW = jnp.zeros(6).at[0:3].set(FW_v_WF[0:3])
+            W_vx_W_FW = Cross.vx(W_v_W_FW)
+            O_Ẋ_W = FW_Ẋ_W = -FW_X_W @ W_vx_W_FW
+
+            O_J̇_WF_I = jnp.zeros(shape=(6, 6 + model.dofs()))
+            O_J̇_WF_I += O_Ẋ_W @ W_J_WL_W @ T
+            O_J̇_WF_I += O_X_W @ W_J̇_WL_W @ T
+            O_J̇_WF_I += O_X_W @ W_J_WL_W @ Ṫ
```
diegoferigo, line 418: Can we move out from the match-case these lines that are shared among all the representations?

tests/test_api_frame.py
```diff
@@ -5,12 +5,12 @@
 
 import jaxsim.api as js
 from jaxsim import VelRepr
+from jaxsim.math.quaternion import Quaternion
 
 from . import utils_idyntree
 
 
 def test_frame_index(jaxsim_models_types: js.model.JaxSimModel):
-
```
diegoferigo, line 13: Nit: can you keep spaces?

tests/test_api_frame.py
```diff
@@ -164,7 +161,132 @@ def test_frame_jacobians(
...
+    # Compute J̇.
+    O_J̇_WF_I = jax.vmap(
+        lambda frame_index: js.frame.jacobian_derivative(
+            model=model, data=data, frame_index=frame_index
+        )
+    )(frame_idxs)
+
+    assert O_J̇_WF_I.shape == (len(frame_names), 6, 6 + model.dofs())
+
+    # Compute O_J̇_WF_I using AD
```
diegoferigo, line 217: leftover?

diegoferigo (approved):
LGTM! Just minor suggestions.

tests/test_api_frame.py
xela-95: Thanks, solved in https://github.com/ami-iit/jaxsim/pull/210/commits/beed30b3a584a538fe5d97f9dad30238defda4ff

tests/test_api_frame.py
xela-95: Thanks, solved in https://github.com/ami-iit/jaxsim/pull/210/commits/beed30b3a584a538fe5d97f9dad30238defda4ff

src/jaxsim/api/frame.py
xela-95: Thanks, solved in https://github.com/ami-iit/jaxsim/pull/210/commits/beed30b3a584a538fe5d97f9dad30238defda4ff

src/jaxsim/api/frame.py
xela-95: Thanks, solved in https://github.com/ami-iit/jaxsim/pull/210/commits/beed30b3a584a538fe5d97f9dad30238defda4ff

flferretti (approved):
That's great @xela-95, thanks a lot! LGTM

PR #212, opened 2024-07-19
Simplify SO3 instantiation and quaternion handling

flferretti:
This pull request simplifies the instantiation of SO3 objects from quaternions in multiple files. It replaces the usage of `jaxlie.SO3.from_quaternion_xyzw` with the more concise `jaxlie.SO3(wxyz=quaternion)`. Additionally, it refactors the reordering of the quaternions from SciPy by using the `scalar_first` option.

### DIFF
--- a/src/jaxsim/math/joint_model.py
+++ b/src/jaxsim/math/joint_model.py
@@ -255,7 +255,7 @@ def supported_joint_motion(
         axis = jnp.array(joint_axis).astype(float).squeeze()
 
         pre_H_suc = jaxlie.SE3.from_rotation(
-            rotation=jaxlie.SO3.from_matrix(Rotation.from_axis_angle(vector=s * axis))
+            rotation=Rotation.from_axis_angle(vector=s * axis)
         )
 
         S = jnp.vstack(jnp.hstack([jnp.zeros(3), axis]))
--- a/src/jaxsim/math/joint_model.py
+++ b/src/jaxsim/math/joint_model.py
@@ -254,8 +254,8 @@ def supported_joint_motion(
         # This is a metadata required by only some joint types.
         axis = jnp.array(joint_axis).astype(float).squeeze()
 
-        pre_H_suc = jaxlie.SE3.from_rotation(
-            rotation=Rotation.from_axis_angle(vector=s * axis)
+        pre_H_suc = jaxlie.SE3.from_matrix(
+            matrix=jnp.eye(4).at[:3, :3].set(Rotation.from_axis_angle(vector=s * axis))
         )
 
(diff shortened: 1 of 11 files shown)


src/jaxsim/math/joint_model.py
```diff
@@ -255,7 +255,7 @@ def compute_R() -> tuple[jtp.Matrix, jtp.Array]:
         axis = jnp.array(joint_axis).astype(float).squeeze()
 
         pre_H_suc = jaxlie.SE3.from_rotation(
-            rotation=jaxlie.SO3.from_matrix(Rotation.from_axis_angle(vector=s * axis))
+            rotation=Rotation.from_axis_angle(vector=s * axis)
```
diegoferigo, line 258: Are you sure that `jaxlie.SE3.from_rotation` accepts a plain jax numpy vector and not a SO3 object? Because `Rotation.from_axis_angle` should return a jax numpy vector.

src/jaxsim/math/rotation.py
```diff
@@ -79,9 +79,11 @@ def theta_is_not_zero(theta_and_v: Tuple[jtp.Float, jtp.Vector]) -> jtp.Matrix:
...
-            false_fun=theta_is_not_zero,
-            operand=(theta, vector),
+        return jaxlie.SO3.from_matrix(
+            jax.lax.cond(
+                pred=(theta == 0.0),
+                true_fun=lambda operand: jnp.eye(3),
+                false_fun=theta_is_not_zero,
+                operand=(theta, vector),
+            )
         )
```
diegoferigo, line 89: Ok just got it, this should not be a `jaxlie.SO3` object but a jax numpy vector. This `math.rotation` helper should not accept nor return any `jaxlie` type.

diegoferigo (changes requested):
Great, thanks! Just a comment to revert an undesired API change.

src/jaxsim/math/rotation.py
flferretti: In https://github.com/ami-iit/jaxsim/pull/212/commits/2e019dd3d6175b35e96a6d12e7b823c4ebf75dc6 I added `.as_matrix()`, so the return type is a `jax.numpy.array`

src/jaxsim/math/joint_model.py
flferretti: Addressed in https://github.com/ami-iit/jaxsim/pull/212/commits/2e019dd3d6175b35e96a6d12e7b823c4ebf75dc6

diegoferigo (approved):
Perfect, it looks good now, thanks!

PR #220, opened 2024-08-19
Fix `JaxSimModelReferences.apply_frame_forces`

flferretti:
----
📚 Documentation preview 📚: https://jaxsim--220.org.readthedocs.build//220/

### DIFF
--- a/src/jaxsim/api/references.py
+++ b/src/jaxsim/api/references.py
@@ -9,7 +9,6 @@ import jax_dataclasses
 import jaxsim.api as js
 import jaxsim.typing as jtp
 from jaxsim import exceptions
-from jaxsim.math import Adjoint
 from jaxsim.utils.tracing import not_tracing
 
 from .common import VelRepr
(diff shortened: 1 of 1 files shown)


diegoferigo (approved):
LGTM! It seems to me that before there were two problems:

1. The returned `references` was wrongly switched to `Inertial` instead of `Body` when the link forces were applied (and the name of the variable `W_f_L_i` should have been `L_f_L_i`).
2. The returned `references` had a different velocity representation than `self`.

Did I get it right?

flferretti:
> 1. The returned `references` was wrongly switched to `Inertial` instead of `Body` when the link forces were applied (and the name of the variable `W_f_L_i` should have been `L_f_L_i`).

I guess it's the other way around, I was converting the frame forces to body and few lines below I was summing them using `Inertial` representation. Keeping everything in inertial, made more sense at this point.

> 2. The returned references had a different velocity representation than self.

This is correct!

There was also a third mistake that had to do with the name of a variable:

```diff
- frame_idx=frame_idxs
+ frame_index=frame_idxs
```

diegoferigo:
> I guess it's the other way around, I was converting the frame forces to body and few lines below I was summing them using Inertial representation. Keeping everything in inertial, made more sense at this point.

Yep what I meant is that the previous logic to express them in the link frames `L` was ok, the problem with that approach is that you were applying them as they were `Inertial` instead of `Body`. Anyway, I agree that keeping everything in inertial as after this PR is more clean.

PR #225, opened 2024-09-05
Add support for 32bit simulations

### LINKED ISSUE #136: Performing a simulation enforcing JAX to run in 32-bit precision overflows after 4.29 seconds

flferretti:
This PR adds experimental support for 32bit simulations. In order to use it, the user can set `JAX_ENABLE_X64=0` 

Closes #136

----
📚 Documentation preview 📚: https://jaxsim--225.org.readthedocs.build//225/


src/jaxsim/api/model.py
```diff
@@ -1931,11 +1932,17 @@ def step(
         ),
     )
 
+    exceptions.raise_if(
+        condition=t0_ns + jnp.array(dt * 1e9).astype(t0_ns.dtype) < t0_ns,
+        exception=OverflowError,
+        msg="The simulation time overflowed the maximum integer value. Consider using x64 by setting `JAX_ENABLE_X64=1`.",
+    )
```
diegoferigo, line 1939: This is good enough for the time being. Though, I am wondering. Fixed step integrators do not need to have a well-formed $t_f$, and I mean that if internally the time overflows, the integrator still works as long as `dt` is finite. What about detecting the overflow after stepping, and reset `t0_ns` to 0?

Something like:

```py
tf_ns = t0_ns + jnp.array(dt * 1e9, dtype=t0_ns.dtype)
tf_ns = jnp.where(tf_ns >= t0_ns, tf_ns, jnp.array(0, dtype=t0_ns))
```

In addition, we might also add a `jax.lax.cond` with a `jax.debug.print` statement only in case of overflow, so that the user is notified in case they use the simulated time in their code.

src/jaxsim/api/model.py
```diff
@@ -1931,11 +1932,17 @@ def step(
...
+
     data_tf = (
         # Store the new state of the model and the new time.
         data.replace(
             state=state_tf,
-            time_ns=t0_ns + jnp.array(dt * 1e9).astype(jnp.uint64),
+            time_ns=t0_ns + jnp.array(dt * 1e9).astype(int),
```
diegoferigo, line 1945:
```suggestion
            time_ns=t0_ns + jnp.array(dt * 1e9, dtype=t0_ns.dtype),
```

src/jaxsim/api/model.py
```diff
@@ -1931,11 +1932,17 @@ def step(
         ),
     )
 
+    exceptions.raise_if(
+        condition=t0_ns + jnp.array(dt * 1e9).astype(t0_ns.dtype) < t0_ns,
```
diegoferigo, line 1936:
```suggestion
        condition=t0_ns + jnp.array(dt * 1e9, dtype=t0_ns.dtype) < t0_ns,
```

diegoferigo (approved):
Looks good. Though, I'm wondering if we can address the overflow in cases when possible.

src/jaxsim/api/model.py
flferretti: Done in https://github.com/ami-iit/jaxsim/pull/225/commits/051704a3690fb977a1f6e666b7846da90be5fb02

src/jaxsim/api/model.py
flferretti: Done in https://github.com/ami-iit/jaxsim/pull/225/commits/051704a3690fb977a1f6e666b7846da90be5fb02

src/jaxsim/api/model.py
flferretti: Solved after discussing f2f. We never raise the OverFlowError, we only print the debug statement.

Addressed in https://github.com/ami-iit/jaxsim/pull/225/commits/59370746f8ece3d931229122366fb60ae329faf5

diegoferigo: If we want, in the future we can perform the detection of overflow _before_ stepping the simulator, and in case of variable-step, raise the `OverFlowError`.

src/jaxsim/api/data.py
```diff
@@ -39,7 +39,7 @@ class JaxSimModelData(common.ModelDataWithVelocityRepresentation):
     contacts_params: jaxsim.rbda.ContactsParams = dataclasses.field(repr=False)
 
     time_ns: jtp.Int = dataclasses.field(
-        default_factory=lambda: jnp.array(0, dtype=jnp.uint64)
+        default_factory=lambda: jnp.array(0, dtype=int)
```
diegoferigo, line 42: I guess that also here we need to select either `uint64` or `uint32` depending on the JAX configuration.

flferretti: Yes, you're right. Thanks a lot! Addressed in https://github.com/ami-iit/jaxsim/pull/225/commits/c58bc190e3337f4cbb74f0b8f863612daeb74a35

PR #226, opened 2024-09-06
Add `pyupgrade` configuration and update to latest `ruff` version

flferretti:
This PR applies the necessary modifications to align with the latest `ruff` version. Moreover, the `pyupgrade` configuration has been added to the linter in order to have a simpler control on the codestyle.

----
📚 Documentation preview 📚: https://jaxsim--226.org.readthedocs.build//226/

### DIFF
--- a/src/jaxsim/api/common.py
+++ b/src/jaxsim/api/common.py
@@ -3,6 +3,7 @@ import contextlib
 import dataclasses
 import enum
 import functools
+from collections.abc import Iterator
 
 import jax
 import jax.numpy as jnp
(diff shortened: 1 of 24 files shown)


src/jaxsim/api/common.py
```diff
@@ -44,7 +43,7 @@ class ModelDataWithVelocityRepresentation(JaxsimDataclass, abc.ABC):
     @contextlib.contextmanager
     def switch_velocity_representation(
         self, velocity_representation: VelRepr
-    ) -> ContextManager[Self]:
+    ) -> contextlib.AbstractContextManager[Self]:
```
diegoferigo, line 46: Mmh this has always been tricky. In practice, this method is a generator that yields an object from another generator. I never managed to get the right configuration that makes the code analysis of PyCharm happy. Is `pyupgrade` happy if the returned type is `Iterator[Self]`?

diegoferigo (approved):

src/jaxsim/api/common.py
flferretti: It should be happy with that as it's not a type checker

flferretti: Addressed in https://github.com/ami-iit/jaxsim/pull/226/commits/2ae3689f2cd7f9616d3b850e5c323c81acc9d9ee

PR #228, opened 2024-09-13
Add support for continuous joints

diegoferigo:
The support in ROD to build programmatically and parse models with continuous joints has been added in https://github.com/ami-iit/rod/pull/40.

A continuous joint, contrarily to revolute joints, does not have `/joint/axis/limit/lower|/joint/axis/limit/upper`. The new version of ROD allows to build a trivial `Limit` object with all attributes set as `None`. In this case, we set the upper and lower limits to the maximum/minimum floating point number in the active precision of JAX (either 64 bits or 32 bits).

----
📚 Documentation preview 📚: https://jaxsim--228.org.readthedocs.build//228/

### DIFF
--- a/src/jaxsim/parsers/rod/parser.py
+++ b/src/jaxsim/parsers/rod/parser.py
@@ -232,15 +232,19 @@ def extract_model_data(
             pose=j.pose.transform() if j.pose is not None else np.eye(4),
             initial_position=0.0,
             position_limit=(
-                (
-                    float(j.axis.limit.lower)
-                    if j.axis is not None and j.axis.limit is not None
-                    else np.finfo(float).min
+                float(
+                    j.axis.limit.lower
+                    if j.axis is not None
+                    and j.axis.limit is not None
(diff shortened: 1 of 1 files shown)


flferretti (approved):
Thanks Diego, LGTM! I just have a curiosity, should these changes be propagated in `JointType`? https://github.com/ami-iit/jaxsim/blob/58b803c426d049232afe74803e6c3db380f31f87/src/jaxsim/parsers/descriptions/joint.py#L15-L20

diegoferigo:
Great question! I don't think we need to do that. The logic of this PR just introduces the support of _parsing_ continous joints (supported in [SDF](http://sdformat.org/spec?ver=1.11&elem=joint#joint_type)). The difference between a continuous joint and a revolute joint is that continuous joints do not have position limits.

This PR basically parses continuous joints and treats them as revolute joints by setting position their limits as the maximum float in JAX precision.

flferretti:
Alright, thanks for the explanation!

PR #229, opened 2024-09-13
Improve sampling random joint positions

diegoferigo:
This PR enhances sampling random joint positions. In particular, it affects how positions from revolute joints are sampled:

- If the position range of the joint is less than $2\pi$, nothing changes.
- If the position range of the joint is larger than $2\pi$ and its range is outside $[-\pi, \\, \pi]$, the position is sampled from $[-\pi, \\, \pi]$.
- If the position range of the joint is larger than $2\pi$ and its lower bound is $s_{\text{min}} > -\pi$, the position is sampled from $[s_{\text{min}}\,\\ s_{\text{min}}+2\pi]$.
- If the position range of the joint is larger than $2\pi$ and its upper bound is $s_{\text{max}} < \pi$, the position is sampled from $[s_{\text{max}} - 2\pi\,\\ s_{\text{max}}]$.

This is particularly useful in presence of either continuous joints or revolute joints with a large position range (#228).

----
📚 Documentation preview 📚: https://jaxsim--229.org.readthedocs.build//229/

### DIFF
--- a/src/jaxsim/api/data.py
+++ b/src/jaxsim/api/data.py
@@ -744,6 +744,13 @@ def random_model_data(
         jtp.FloatLike | Sequence[jtp.FloatLike],
         jtp.FloatLike | Sequence[jtp.FloatLike],
     ] = ((-1, -1, 0.5), 1.0),
+    joint_pos_bounds: (
+        tuple[
+            jtp.FloatLike | Sequence[jtp.FloatLike],
+            jtp.FloatLike | Sequence[jtp.FloatLike],
+        ]
+        | None
+    ) = None,
     base_vel_lin_bounds: tuple[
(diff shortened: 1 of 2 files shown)


flferretti (approved):
Thanks Diego, it looks good! What about porting the new logic inside `js.joint.random_joint_positions`? So we can use this feature also there

diegoferigo:
> What about porting the new logic inside js.joint.random_joint_positions? So we can use this feature also there

I didn't get your point. The logic is already in that function.

flferretti:
> > What about porting the new logic inside js.joint.random_joint_positions? So we can use this feature also there
> 
> I didn't get your point. The logic is already in that function.

My bad, sorry :)

PR #230, opened 2024-09-13
Add support for sampling `JaxSimModelData` objects with random base orientation

diegoferigo:
The logic uses a sequence of three Euler angles and allows to specify the sequence with a string compatible with [`scipy.spatial.transform.Rotation.from_euler()`](https://docs.scipy.org/doc/scipy/reference/generated/scipy.spatial.transform.Rotation.from_euler.html).

----
📚 Documentation preview 📚: https://jaxsim--230.org.readthedocs.build//230/

### DIFF
--- a/src/jaxsim/api/data.py
+++ b/src/jaxsim/api/data.py
@@ -7,9 +7,9 @@ from collections.abc import Sequence
 import jax
 import jax.numpy as jnp
 import jax_dataclasses
-import jaxlie
 
 import jaxsim.api as js
+import jaxsim.math
 import jaxsim.rbda
 import jaxsim.typing as jtp
 from jaxsim.rbda.contacts.soft import SoftContacts
@@ -195,10 +195,9 @@ class JaxSimModelData(common.ModelDataWithVelocityRepresentation):
         else:
             contacts_params = model.contact_model.parameters
 
-        W_H_B = jaxlie.SE3.from_rotation_and_translation(
-            translation=base_position,
-            rotation=jaxlie.SO3(wxyz=base_quaternion),
-        ).as_matrix()
+        W_H_B = jaxsim.math.Transform.from_quaternion_and_translation(
+            translation=base_position, quaternion=base_quaternion
+        )
 
         v_WB = JaxSimModelData.other_representation_to_inertial(
             array=jnp.hstack([base_linear_velocity, base_angular_velocity]),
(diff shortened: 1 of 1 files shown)


src/jaxsim/api/data.py
```diff
@@ -819,9 +833,14 @@ def random_model_data(
...
-            *jax.random.uniform(key=k2, shape=(3,), minval=0, maxval=2 * jnp.pi)
-        ).wxyz
+        physics_model_state.base_quaternion = jaxsim.math.Quaternion.to_wxyz(
+            xyzw=jax.scipy.spatial.transform.Rotation.from_euler(
+                seq=base_rpy_seq,
+                angles=jax.random.uniform(
+                    key=k2, shape=(3,), minval=rpy_min, maxval=rpy_max
+                ),
+            ).as_quat()
+        )
```
flferretti, line 843: Maybe we should add `scalar_first=True` in order to set the quaternion as **wxyz** 
```suggestion
        physics_model_state.base_quaternion = jax.scipy.spatial.transform.Rotation.from_euler(
            seq=base_rpy_seq,
            angles=jax.random.uniform(
                key=k2, shape=(3,), minval=rpy_min, maxval=rpy_max
            ),
        ).as_quat(scalar_first=True)
```

flferretti (changes requested):

src/jaxsim/api/data.py
diegoferigo: The JAX implementation, at the time of writing, does not have `scalar_first` as in official SciPy:

https://github.com/google/jax/blob/427a490d2b4147c41c84560792936d693d3c921a/jax/_src/scipy/spatial/transform.py#L170-L175

flferretti: Oh alright, then what about something like:

```python
Q = jax.scipy.spatial.transform.Rotation.from_euler(
            seq=base_rpy_seq,
            angles=jax.random.uniform(
                key=k2, shape=(3,), minval=rpy_min, maxval=rpy_max
            ),
        ).as_quat()
Q = jnp.roll(Q, shift=-1)
```

flferretti (approved):

src/jaxsim/api/data.py
diegoferigo: When I read the roll code, to me is never clear the direction of rolling. I prefer much more the more verbose call `Quaternion.to_wxyz(xyzw=...)` that at least is explicit on what it is doing. Considering the runtime, there should not be any difference.

flferretti: Alright, sounds good

PR #232, opened 2024-09-17
Fix `jaxsim.api.model.link_contact_forces`

diegoferigo:
This function was a leftover from when we updated the velocity representation of the model Jacobian to not use $B$ and $B[W]$ in body-fixed and mixed in favor of $L$ and $L[W]$ (https://github.com/ami-iit/jaxsim/pull/167).

Luckily this method was not used elsewhere in the code -- and probably this is why this leftover got under our radar. However, there was a script on which @xela-95 and us worked that was calling this function. Everything was correct it inerital-fixed, but things were not correct in other velocity representations.

----
📚 Documentation preview 📚: https://jaxsim--232.org.readthedocs.build//232/

### DIFF
--- a/src/jaxsim/api/model.py
+++ b/src/jaxsim/api/model.py
@@ -1747,14 +1747,18 @@ def link_contact_forces(
         data: The data of the considered model.
 
     Returns:
-        A (nL, 6) array containing the stacked 6D contact forces of the links,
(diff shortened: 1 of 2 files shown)


flferretti (approved):
Thanks a lot Diego! LGTM

PR #233, opened 2024-09-17
Account for joint torques in the rigid contact models

flferretti:
This pull request includes several updates to the contact models in the codebase. The `jaxsim.api.contact.collidable_point_dynamics` function now takes into account the applied torques in the rigid contact models.  The `RelaxedRigidContacts` class have been updated to include docstrings and a deprecation warning on `jax.tree_map` has been fixed

----
📚 Documentation preview 📚: https://jaxsim--233.org.readthedocs.build//233/

### DIFF
--- a/src/jaxsim/rbda/contacts/relaxed_rigid.py
+++ b/src/jaxsim/rbda/contacts/relaxed_rigid.py
@@ -202,7 +202,7 @@ class RelaxedRigidContacts(ContactModel):
             link_forces:
                 Optional `(n_links, 6)` matrix of external forces acting on the links,
                 expressed in the same representation of data.
-            joint_force_reference:
+            joint_force_references:
                 Optional `(n_joints,)` vector of joint forces.
 
         Returns:
@@ -226,6 +226,7 @@ class RelaxedRigidContacts(ContactModel):
             data=data,
             velocity_representation=data.velocity_representation,
             link_forces=link_forces,
+            joint_force_references=joint_force_references,
         )
 
         def _detect_contact(x: jtp.Array, y: jtp.Array, z: jtp.Array) -> jtp.Array:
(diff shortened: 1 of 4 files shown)


src/jaxsim/api/kin_dyn_parameters.py
```diff
@@ -98,9 +98,7 @@ def build(model_description: ModelDescription) -> KynDynParameters:
         ]
 
         # Create a vectorized object of link parameters.
-        link_parameters = jax.tree_util.tree_map(
-            lambda *l: jnp.stack(l), *link_parameters_list
-        )
+        link_parameters = jax.tree.map(lambda *l: jnp.stack(l), *link_parameters_list)
```
diegoferigo, line 101: Can you please check what is the first JAX release that supports this? I suspect we need to bump the version we currently support.

src/jaxsim/rbda/contacts/relaxed_rigid.py
```diff
@@ -189,23 +189,48 @@ def compute_contact_forces(
...
         link_forces = (
             link_forces
             if link_forces is not None
             else jnp.zeros((model.number_of_links(), 6))
         )
 
+        joint_force_references = (
+            joint_force_references
+            if joint_force_references is not None
+            else jnp.zeros((model.number_of_joints(),))
```
diegoferigo, line 221:
```suggestion
            else jnp.zeros(model.number_of_joints())
```

diegoferigo (approved):
Looks good! Thanks for the fix. If you think it's worth, feel free to split the update of tree map calls into another PR.

src/jaxsim/rbda/contacts/relaxed_rigid.py
flferretti: Addressed in https://github.com/ami-iit/jaxsim/pull/233/commits/64bef4dfb446ff8103d67924f6cee23da010896e

PR #236, opened 2024-09-19
Simplify CRBA by merging two `jax.lax.cond` branches

diegoferigo:
This change doesn't significantly affect speed, but it simplifies the logic that implements a nested while loop with a `jax.lax.scan` -- necessary to support reverse-mode AD.

I have the impression that we can remove entirely this while loop by slicing the mass matrix with $\nu(i)$, that defines the bodies supported by joint $i$, as @flferretti is trying in #231, but I didn't investigate further in this direction. 

----
📚 Documentation preview 📚: https://jaxsim--236.org.readthedocs.build//236/


flferretti (approved):
Thanks a lot Diego, LGTM!

PR #237, opened 2024-09-19
Rigid contacts speedup improvements

xela-95:
This PR updates the RigidContacts class removing redundant calls to get kin-dyn quantities like contact jacobians, system velocity, etc.

It also exposes the QP solver tolerance.

----
📚 Documentation preview 📚: https://jaxsim--237.org.readthedocs.build//237/

### DIFF
--- a/src/jaxsim/rbda/contacts/rigid.py
+++ b/src/jaxsim/rbda/contacts/rigid.py
@@ -215,6 +215,7 @@ class RigidContacts(ContactModel):
         link_forces: jtp.MatrixLike | None = None,
         joint_force_references: jtp.VectorLike | None = None,
         regularization_term: jtp.FloatLike = 1e-6,
+        solver_tol: jtp.FloatLike = 1e-3,
     ) -> tuple[jtp.Vector, tuple[Any, ...]]:
         """
         Compute the contact forces.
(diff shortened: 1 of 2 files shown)


src/jaxsim/rbda/contacts/rigid.py
```diff
@@ -194,7 +195,7 @@ def impact_velocity(
                 return nu_post
 
         with data.switch_velocity_representation(VelRepr.Mixed):
-            BW_ν_pre_impact = data.generalized_velocity()
+            BW_ν_pre_impact = BW_nu
```
diegoferigo, line 198: Extracting $\boldsymbol{\nu}$ from `data` has almost zero overhead, I'm not sure that passing it from the outside is worth. This change introduces the risk that the caller (maybe in user code) passes a velocity unaligned with `data`. What do you think?

diegoferigo (changes requested):
Early feedback.

src/jaxsim/rbda/contacts/rigid.py
xela-95: Ok, I agree that this can introduce more risks than performance advantages. I will remove this change in the next commit.

xela-95: Done in 2af4b8b8c07f45e3e64689d7a2799a097d13a609

src/jaxsim/rbda/contacts/rigid.py
```diff
@@ -257,6 +259,12 @@ def compute_contact_forces(
             M = js.model.free_floating_mass_matrix(model=model, data=data)
             J_WC = js.contact.jacobian(model=model, data=data)
             W_H_C = js.contact.transforms(model=model, data=data)
+            J̇_WC_BW = js.contact.jacobian_derivative(
+                model=model,
+                data=data,
+            )
```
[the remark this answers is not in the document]
xela-95: Done in 2af4b8b8c07f45e3e64689d7a2799a097d13a609

xela-95:
Addressed review comments and rebased on main.

flferretti (approved):
That's great! Thanks @xela-95, LGTM

diegoferigo (approved):
Great! Thanks @xela-95 for the speedup improvement.

PR #238, opened 2024-09-19
Remove `is_urdf` argument when building the model

flferretti:
This PR removes the `is_urdf` argument, as it can get automatically detected by `rod`. Furthermore, it enables using a `rod.Sdf` instance to build a `JaxSimModel`.

> [!Warning]
> Requires https://github.com/ami-iit/rod/pull/42

----
📚 Documentation preview 📚: https://jaxsim--238.org.readthedocs.build//238/

### DIFF
--- a/src/jaxsim/mujoco/loaders.py
+++ b/src/jaxsim/mujoco/loaders.py
@@ -25,7 +25,7 @@ def load_rod_model(
 
     Args:
         model_description: The URDF/SDF file or ROD model to load.
-        is_urdf: Whether the model description is a URDF file.
+        is_urdf: Whether to force parsing the model description as a URDF file.
         model_name: The name of the model to load from the resource.
 
     Returns:
@@ -558,7 +558,6 @@ class UrdfToMjcf:
         # Get the ROD model.
         rod_model = load_rod_model(
             model_description=urdf,
-            is_urdf=True,
             model_name=model_name,
         )
 
@@ -605,7 +604,6 @@ class SdfToMjcf:
         # Get the ROD model.
         rod_model = load_rod_model(
             model_description=sdf,
-            is_urdf=False,
             model_name=model_name,
         )
 
--- a/src/jaxsim/mujoco/loaders.py
+++ b/src/jaxsim/mujoco/loaders.py
@@ -558,6 +558,7 @@ class UrdfToMjcf:
         # Get the ROD model.
         rod_model = load_rod_model(
             model_description=urdf,
+            is_urdf=True,
             model_name=model_name,
         )
 
@@ -604,6 +605,7 @@ class SdfToMjcf:
         # Get the ROD model.
         rod_model = load_rod_model(
             model_description=sdf,
+            is_urdf=False,
             model_name=model_name,
         )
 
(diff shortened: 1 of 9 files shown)


diegoferigo:
I just released rod [`v0.3.2`](https://github.com/ami-iit/rod/releases/tag/v0.3.2), can you please bump the dependency? CI should pass as soon as it is deployed to PyPI (maybe in ~10 mins).

diegoferigo:
@flferretti also in `environment.yml` :)

flferretti:
> I just released rod [`v0.3.2`](https://github.com/ami-iit/rod/releases/tag/v0.3.2), can you please bump the dependency? CI should pass as soon as it is deployed to PyPI (maybe in ~10 mins).

Thanks for the release! I made the requested changes on the requirements, it should be okay now

flferretti:
RTD action will fail until `rod==0.3.3` will be released on conda-forge. I triggered https://github.com/conda-forge/rod-feedstock/pull/12

flferretti:
As discussed f2f, I added the possibility to force parsing the resource as URDF.

src/jaxsim/mujoco/loaders.py
```diff
@@ -558,7 +558,6 @@ def convert(
         # Get the ROD model.
         rod_model = load_rod_model(
             model_description=urdf,
-            is_urdf=True,
```
diegoferigo, line 561: Since we assume that the model is URDF, this should fail is `urdf` is a SDF string. I'd leave this flag here in this specific case.

src/jaxsim/parsers/rod/parser.py
```diff
@@ -41,30 +41,37 @@ def extract_model_data(
...
-
-        # Assume the SDF resource has only one model, or the desired model name is given.
-        sdf_models = {m.name: m for m in sdf_element.models()}
-        sdf_model = (
-            sdf_element.models()[0] if len(sdf_models) == 1 else sdf_models[model_name]
-        )
+    match model_description:
+        case rod.Model():
+            sdf_model = model_description
+        case rod.Sdf(None):
```
diegoferigo, line 59: This case was not originally part of the logic. It's a good addition having it here but there's something missing:

- Update the typing of the argument.
- Apply the logic to extract the right model used in the last case block also here (possibly with no code duplication).

diegoferigo (changes requested):
Last fix since the changes introduced a minor change of behavior in a new considered case.

src/jaxsim/parsers/rod/parser.py
flferretti: I agree with that! Addressed in https://github.com/ami-iit/jaxsim/pull/238/commits/24e91a78a0c374826827665e72277fecfce7ee41

src/jaxsim/mujoco/loaders.py
flferretti: Addressed in https://github.com/ami-iit/jaxsim/pull/238/commits/24e91a78a0c374826827665e72277fecfce7ee41

diegoferigo (approved):
Great! Thanks @flferretti for addressing feedback.

PR #240, opened 2024-09-20
Fix references representation in `system_velocity_dynamics`

xela-95:
In `system_velocity_dynamics` we create a references instance with the input velocity representation of data, then we call `js.contact.collidable_point_dynamics` inside a context manager in which data representation has been switched to Inertial, but in this case the references object has not been switched accordingly, with the result that in the cases of input data representation being `Body` or `Mixed` the link forces passed to `js.contact.collidable_point_dynamics` could be wrong.

Let me know what do you think :)

### DIFF
--- a/src/jaxsim/api/ode.py
+++ b/src/jaxsim/api/ode.py
@@ -145,7 +145,10 @@ def system_velocity_dynamics(
 
         # Compute the 6D forces W_f ∈ ℝ^{n_c × 6} applied to each collidable point
         # along with contact-specific auxiliary states.
-        with data.switch_velocity_representation(VelRepr.Inertial):
+        with (
+            data.switch_velocity_representation(VelRepr.Inertial),
+            references.switch_velocity_representation(VelRepr.Inertial),
+        ):
             W_f_Ci, aux_data = js.contact.collidable_point_dynamics(
                 model=model,
                 data=data,


flferretti (approved):
Thanks a lot @xela-95, LGTM!

diegoferigo (approved):
Good catch @xela-95! Probably it went under our radar in one of the latest changes of this function.

Just yesterday I found a problem in `js.contact.collidable_point_dynamics` when the representation of data is not `Inertial`. I still need to push it, though. If this is the use-case in any of your scripts, let me know so that I prioritize the PR.

xela-95:
> Just yesterday I found a problem in `js.contact.collidable_point_dynamics` when the representation of data is not `Inertial`. I still need to push it, though. If this is the use-case in any of your scripts, let me know so that I prioritize the PR.

Ah ok! @diegoferigo if you want to push a PR even in draft it will be great for me since I'm testing the behavior of contacts under different conditions :)

diegoferigo:
> > Just yesterday I found a problem in `js.contact.collidable_point_dynamics` when the representation of data is not `Inertial`. I still need to push it, though. If this is the use-case in any of your scripts, let me know so that I prioritize the PR.
> 
> Ah ok! @diegoferigo if you want to push a PR even in draft it will be great for me since I'm testing the behavior of contacts under different conditions :)

Sure, done in https://github.com/ami-iit/jaxsim/pull/241.

PR #242, opened 2024-09-23
Fix `jaxsim.api.model.link_contact_forces`

xela-95:
In this PR I added to the `jaxsim.api.model.link_contact_forces` API the external link forces and joint torques as additional input arguments. These are then passed to `jaxsim.api.contact.collidable_point_forces`. In this way, the contact forces computed with the rigid contact model (but this applies also to relaxed-rigid ones I think) will consider all the forces acting on the model.

I faced an issue when I was performing simulations with @flferretti using rigid contacts, and we noted that the contact forces given by   `jaxsim.api.model.link_contact_forces` were only related to the weight of the model and were not considering other forces potentially acting on the model. 

The computation of the contact forces used inside `model.step` was already correct: through `ode.system_velocity_dynamics` it calls `contact.collidable_point_dynamics` with the correct arguments, while `contact.collidable_point_forces` was not aligned.

Let me know what do you think of this :)

----
📚 Documentation preview 📚: https://jaxsim--242.org.readthedocs.build//242/

### DIFF
--- a/src/jaxsim/api/contact.py
+++ b/src/jaxsim/api/contact.py
@@ -93,7 +93,10 @@ def collidable_point_velocities(
 
 @jax.jit
 def collidable_point_forces(
-    model: js.model.JaxSimModel, data: js.data.JaxSimModelData
+    model: js.model.JaxSimModel,
+    data: js.data.JaxSimModelData,
+    link_forces: jtp.MatrixLike | None = None,
+    joint_force_references: jtp.VectorLike | None = None,
 ) -> jtp.Matrix:
     """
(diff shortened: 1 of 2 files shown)


xela-95:
Also I'm not sure where it is more appropriate to validate the input arguments in the chain of function calls, for now I just check the input link forces and joint torques in `jaxsim.api.model.link_contact_forces`.

flferretti (approved):
Thanks a lot @xela-95 for the fix, LGTM

diegoferigo (approved):
Thanks @xela-95, LGTM!

PR #243, opened 2024-09-24
Allow calls to `logging` in `jaxsim.exceptions`

flferretti:
This PR is a proposal to adapt the `jaxsim.exceptions` so that it can also work with calls to `jaxsim.logging` or `logging`. Eventually, we could think of just separating the two functionalities or renaming the module.

----
📚 Documentation preview 📚: https://jaxsim--243.org.readthedocs.build//243/

### DIFF
--- a/src/jaxsim/exceptions.py
+++ b/src/jaxsim/exceptions.py
@@ -1,22 +1,22 @@
 import jax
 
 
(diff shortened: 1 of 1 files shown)


flferretti:
Closing as this is no longer needed

PR #245, opened 2024-09-26
Refactor `SoftContacts` algorithm

diegoferigo:
I realized that we can greatly simplify the algorithm by removing all conditional branches. It might run slightly slower since it has to compute a few more variables, but the code becomes much more readable and easy to debug in case of problems.

I tried to run some simulations either with spheres (that are approximated with 250 points) and humanoid robots free falling on terrain, and performance are almost the same, in few cases I get average execution time couple of % points slower, which is more than acceptable.

Furthermore, I generalized some soft contact resources for something else that will land soon :smiley: 

----
📚 Documentation preview 📚: https://jaxsim--245.org.readthedocs.build//245/

### DIFF
--- a/src/jaxsim/rbda/contacts/relaxed_rigid.py
+++ b/src/jaxsim/rbda/contacts/relaxed_rigid.py
@@ -1,7 +1,7 @@
 from __future__ import annotations
 
 import dataclasses
-from typing import Any, Type
+from typing import Any
 
 import jax
 import jax.numpy as jnp
@@ -111,7 +111,7 @@ class RelaxedRigidContactsParams(ContactsParams):
 
     @classmethod
     def build(
-        cls: Type[Self],
+        cls: type[Self],
         *,
         time_constant: jtp.FloatLike | None = None,
         damping_coefficient: jtp.FloatLike | None = None,
@@ -163,13 +163,13 @@ class RelaxedRigidContactsState(ContactsState):
         return hash(self) == hash(other)
 
     @classmethod
-    def build(cls: Type[Self]) -> Self:
+    def build(cls: type[Self]) -> Self:
         """Create a `RelaxedRigidContactsState` instance"""
 
         return RelaxedRigidContactsState()
 
     @classmethod
-    def zero(cls: Type[Self], *, model: js.model.JaxSimModel) -> Self:
+    def zero(cls: type[Self], *, model: js.model.JaxSimModel) -> Self:
         """Build a zero `RelaxedRigidContactsState` instance from a `JaxSimModel`."""
 
         return RelaxedRigidContactsState.build()
--- a/src/jaxsim/rbda/contacts/relaxed_rigid.py
+++ b/src/jaxsim/rbda/contacts/relaxed_rigid.py
@@ -166,13 +166,13 @@ class RelaxedRigidContactsState(ContactsState):
     def build(cls: type[Self]) -> Self:
         """Create a `RelaxedRigidContactsState` instance"""
 
-        return RelaxedRigidContactsState()
+        return cls()
 
     @classmethod
-    def zero(cls: type[Self], *, model: js.model.JaxSimModel) -> Self:
+    def zero(cls: type[Self]) -> Self:
         """Build a zero `RelaxedRigidContactsState` instance from a `JaxSimModel`."""
 
-        return RelaxedRigidContactsState.build()
+        return cls.build()
 
     def valid(self, *, model: js.model.JaxSimModel) -> jtp.BoolLike:
         return True
--- a/src/jaxsim/rbda/contacts/relaxed_rigid.py
+++ b/src/jaxsim/rbda/contacts/relaxed_rigid.py
@@ -1,7 +1,7 @@
 from __future__ import annotations
 
 import dataclasses
-from typing import Any
+from typing import Any, Type
 
 import jax
 import jax.numpy as jnp
@@ -15,6 +15,11 @@ from jaxsim.math import Adjoint
 from jaxsim.terrain.terrain import FlatTerrain, Terrain
 
 from .common import ContactModel, ContactsParams, ContactsState
+
+try:
+    from typing import Self
+except ImportError:
+    from typing_extensions import Self
 
 
 @jax_dataclasses.pytree_dataclass
@@ -106,7 +111,8 @@ class RelaxedRigidContactsParams(ContactsParams):
 
     @classmethod
     def build(
-        cls,
+        cls: Type[Self],
+        *,
         time_constant: jtp.FloatLike | None = None,
         damping_coefficient: jtp.FloatLike | None = None,
         d_min: jtp.FloatLike | None = None,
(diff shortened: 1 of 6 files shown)


src/jaxsim/rbda/contacts/relaxed_rigid.py
```diff
@@ -155,18 +162,19 @@ class RelaxedRigidContactsState(ContactsState):
     def __eq__(self, other: RelaxedRigidContactsState) -> bool:
         return hash(self) == hash(other)
 
-    @staticmethod
-    def build() -> RelaxedRigidContactsState:
+    @classmethod
+    def build(cls: type[Self]) -> Self:
         """Create a `RelaxedRigidContactsState` instance"""
 
         return RelaxedRigidContactsState()
```
flferretti, line 169: At this point we could directly use `cls`
```suggestion
        return cls()
```

src/jaxsim/rbda/contacts/relaxed_rigid.py
```diff
@@ -155,18 +162,19 @@ class RelaxedRigidContactsState(ContactsState):
...
 
-    @staticmethod
-    def zero(model: js.model.JaxSimModel) -> RelaxedRigidContactsState:
+    @classmethod
+    def zero(cls: type[Self], *, model: js.model.JaxSimModel) -> Self:
         """Build a zero `RelaxedRigidContactsState` instance from a `JaxSimModel`."""
+
         return RelaxedRigidContactsState.build()
```
flferretti, line 175: Same as above
```suggestion
        return cls.build()
```

src/jaxsim/rbda/contacts/relaxed_rigid.py
```diff
@@ -155,18 +162,19 @@ class RelaxedRigidContactsState(ContactsState):
...
+    @classmethod
+    def build(cls: type[Self]) -> Self:
         """Create a `RelaxedRigidContactsState` instance"""
 
         return RelaxedRigidContactsState()
 
-    @staticmethod
-    def zero(model: js.model.JaxSimModel) -> RelaxedRigidContactsState:
+    @classmethod
+    def zero(cls: type[Self], *, model: js.model.JaxSimModel) -> Self:
```
flferretti, line 172: We can actually remove `model` since it is not used. The signature shouldn't change with respect to the abstract class
```suggestion
    def zero(cls: type[Self]) -> Self:
```

src/jaxsim/rbda/contacts/rigid.py
```diff
@@ -50,19 +55,22 @@ def __eq__(self, other: RigidContactsParams) -> bool:
...
+        cls: type[Self],
+        *,
         mu: jtp.FloatLike | None = None,
         K: jtp.FloatLike | None = None,
         D: jtp.FloatLike | None = None,
-    ) -> RigidContactsParams:
+    ) -> Self:
         """Create a `RigidContactParams` instance"""
+
         return RigidContactsParams(
```
flferretti, line 66: Same as above
```suggestion
        return cls(
```

src/jaxsim/rbda/contacts/rigid.py
```diff
@@ -77,18 +85,19 @@ class RigidContactsState(ContactsState):
     def __eq__(self, other: RigidContactsState) -> bool:
         return hash(self) == hash(other)
 
-    @staticmethod
-    def build(**kwargs) -> RigidContactsState:
+    @classmethod
+    def build(cls: type[Self]) -> Self:
         """Create a `RigidContactsState` instance"""
 
         return RigidContactsState()
```
flferretti, line 92: Same as above
```suggestion
        return cls()
```

src/jaxsim/rbda/contacts/rigid.py
```diff
@@ -77,18 +85,19 @@ class RigidContactsState(ContactsState):
...
 
-    @staticmethod
-    def zero(**kwargs) -> RigidContactsState:
+    @classmethod
+    def zero(cls: type[Self]) -> Self:
         """Build a zero `RigidContactsState` instance from a `JaxSimModel`."""
+
         return RigidContactsState.build()
```
flferretti, line 98: Same as above
```suggestion
        return cls.build()
```

src/jaxsim/rbda/contacts/soft.py
```diff
@@ -126,21 +165,25 @@ def build_default_from_jaxsim_model(
...
-        )
+        return jnp.hstack(
+            [
+                self.K >= 0.0,
+                self.D >= 0,
+                self.mu >= 0.0,
+                self.p >= 0.0,
+                self.q >= 0.0,
+            ]
+        ).all()
```
flferretti, line 186: Can we just use the built-in `all`?
```suggestion
        return all(
            [
                self.K >= 0.0,
                self.D >= 0,
                self.mu >= 0.0,
                self.p >= 0.0,
                self.q >= 0.0,
            ]
        )
```

src/jaxsim/rbda/contacts/soft.py
```diff
@@ -126,21 +165,25 @@ def build_default_from_jaxsim_model(
...
 
-        return (
-            jnp.all(self.K >= 0.0)
-            and jnp.all(self.D >= 0.0)
-            and jnp.all(self.mu >= 0.0)
-        )
+        return jnp.hstack(
+            [
+                self.K >= 0.0,
+                self.D >= 0,
```
flferretti, line 181:
```suggestion
                self.D >= 0.0,
```

src/jaxsim/rbda/contacts/soft.py
```diff
@@ -157,179 +200,161 @@ class SoftContacts(ContactModel):
...
+        # 1: sticking
+        # 2: no contact
+        contact_status = sticking.astype(int)
+        contact_status += (δ <= 0).astype(int)
+
+        # Select the right material deformation rate depending on the contact status.
+        ṁ = jax.lax.switch(
+            index=contact_status,
+            branches=(lambda: ṁ_slipping, lambda: ṁ_sticking, lambda: ṁ_no_contact),
+        )
```
flferretti, line 316: We could also use `jax.lax.select_n` if it is more sound to you
```suggestion
        ṁ = jax.lax.select_n(
            ṁ_slipping, ṁ_sticking, ṁ_no_contact,
            which=contact_status,
        )
```

src/jaxsim/rbda/contacts/soft.py
```diff
@@ -157,179 +200,161 @@ class SoftContacts(ContactModel):
...
+        # Compute the penetration depth normal to the terrain.
+        δ = jnp.maximum(0.0, jnp.dot(h, n̂))
+
+        # Compute the penetration normal velocity.
+        δ̇ = -jnp.dot(W_ṗ_C, n̂)
+
+        # Enforce the penetration rate to be zero when the penetration depth is zero.
+        δ̇ = jnp.where(δ > 0, δ̇, 0.0)
+
+        return δ, δ̇, n̂
```
flferretti, line 357: Since we are using this in various parts of the code, we might consider putting it in `jaxsim.api.contact`

flferretti (changes requested):
Thanks a lot Diego! I just added some minor comments

src/jaxsim/rbda/contacts/soft.py
diegoferigo: If we use all, I'm not sure we can use the resulting bool to raise jit-compatible exceptions. We should double check this sooner or later and align all the `valid` methods scattered around.

flferretti (approved):
Thanks a lot for addressing the review!

xela-95 (approved):
Great @diegoferigo! It looks good to me 🚀

PR #246, opened 2024-09-27
Allow `Rigid*ContactsState` to accept kwargs

flferretti:
This fixes a regression introduced from a suggestion I made in https://github.com/ami-iit/jaxsim/pull/245. When the zero-ed ODEState is created in https://github.com/ami-iit/jaxsim/blob/a4d14766766ba838ea784ddf15df6feb36465164/src/jaxsim/api/ode_data.py#L250-L252
this raises an error in rigid contact models:

```python
TypeError: RelaxedRigidContactsState.zero() got an unexpected keyword argument 'model'
```

----
📚 Documentation preview 📚: https://jaxsim--246.org.readthedocs.build//246/

### DIFF
--- a/src/jaxsim/rbda/contacts/relaxed_rigid.py
+++ b/src/jaxsim/rbda/contacts/relaxed_rigid.py
@@ -169,7 +169,7 @@ class RelaxedRigidContactsState(ContactsState):
         return cls()
 
     @classmethod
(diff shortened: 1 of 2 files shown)


diegoferigo (approved):
Thanks @flferretti for spotting this regression!

diegoferigo (approved):

PR #247, opened 2024-09-27
Update `ContactModel.compute_contact_forces` APIs

diegoferigo:
The new contact models that have been introduced recently need to consider the whole system dynamics to compute the contact forces. This was not necessary in the original `SoftContacts` model. However, this is just a simplification specific to the `SoftContacts` model implementation, and it is the main reason why they can run so fast compared to the new models.

I believe that the most generic APIs take `model` and `data` as input arguments instead of `position` and `velocity`. This also aligns the disparity in which the `position|velocity` arguments of `SoftContacts` were the individual ones, while for the other contact models were the vmapped ones. Furthermore, it removes the ambiguity for the other contact models in which the passed  `position|velocity` could differ from the kinematics that can be computed from `model|data`.

----
📚 Documentation preview 📚: https://jaxsim--247.org.readthedocs.build//247/

### DIFF
--- a/src/jaxsim/api/data.py
+++ b/src/jaxsim/api/data.py
@@ -185,20 +185,6 @@ class JaxSimModelData(common.ModelDataWithVelocityRepresentation):
             )
         )
 
-        contacts_params = (
-            contacts_params
-            if contacts_params is not None
-            else (
-                model.contact_model.parameters
-                if not isinstance(
-                    model.contact_model, jaxsim.rbda.contacts.SoftContacts
-                )
-                else js.contact.estimate_good_soft_contacts_parameters(
-                    model=model, standard_gravity=standard_gravity
-                )
-            )
-        )
-
         W_H_B = jaxsim.math.Transform.from_quaternion_and_translation(
             translation=base_position, quaternion=base_quaternion
         )
@@ -227,6 +213,15 @@ class JaxSimModelData(common.ModelDataWithVelocityRepresentation):
 
         if not ode_state.valid(model=model):
             raise ValueError(ode_state)
+
+        if contacts_params is None:
+
+            if isinstance(model.contact_model, jaxsim.rbda.contacts.SoftContacts):
+                contacts_params = js.contact.estimate_good_soft_contacts_parameters(
+                    model=model, standard_gravity=standard_gravity
+                )
+            else:
+                contacts_params = model.contact_model.parameters
 
         return JaxSimModelData(
             time_ns=time_ns,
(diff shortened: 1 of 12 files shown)


flferretti (approved):
Thanks Diego, LGTM!

xela-95 (approved):
Thanks @diegoferigo!! LGTM!

src/jaxsim/api/ode_data.py
```diff
@@ -165,8 +167,11 @@ def build_from_jaxsim_model(
 
         # Get the contact model from the `JaxSimModel`.
         match model.contact_model:
+
```
flferretti, line 170: This is a bit out of scope, but what about using `ruff format` instead of `black`, so we can get coherent spaces across the codesource?

src/jaxsim/api/data.py
```diff
@@ -185,16 +185,19 @@ def build(
...
+            if contacts_params is not None
+            else (
+                model.contact_model.parameters
+                if not isinstance(
+                    model.contact_model, jaxsim.rbda.contacts.SoftContacts
+                )
                 else js.contact.estimate_good_soft_contacts_parameters(
                     model=model, standard_gravity=standard_gravity
                 )
             )
```
flferretti, line 199: Maybe we could make the logic a bit more linear instead of nesting two ternary operators. Note that the suggestion has to be reformatted as I cannot comment on deleted lines
```suggestion
if contacts_params is None:
    if isinstance(model.contact_model, jaxsim.rbda.contacts.SoftContacts):
        contacts_params = js.contact.estimate_good_soft_contacts_parameters(
            model=model, standard_gravity=standard_gravity
        )
    else:
        contacts_params = model.contact_model.parameters
```

flferretti (changes requested):

src/jaxsim/api/data.py
diegoferigo: Done in https://github.com/ami-iit/jaxsim/pull/247/commits/4fc99ea7ac62fe3b54f08471261b00dd730f2250

src/jaxsim/api/ode_data.py
diegoferigo: Mmh how does ruff format play with VSCode? AFAIK the default formatter is black. I need to double check in PyCharm.

flferretti: It's only about installing the `ruff` extension and choosing it from the dropdown menu
 
image

flferretti: It may be useful to checkout https://docs.astral.sh/ruff/formatter/black/

diegoferigo: Ok, we can discuss this f2f next week before starting with the possible migration.

flferretti (approved):
Thanks for addressing the review! This improves a lot the readability and the logic of that part of the codebase

PR #252, opened 2024-10-03
Refactor `integrators` module and  redefine `JaxSimModel` class attributes

flferretti:
This pull request includes several changes to the integrators logic and API. The `JaxSimModel` class now incorporates `integrator` and `dt` as static attributes. In order to do this, it was necessary to refactor the `jaxsim.integrators` module in order to simplify the logic by removing the `init` method.

The integrators logic is simplified by avoiding the use of `integrator_state` to pass around variables and making `wrap_system_dynamics_for_integration` accept only the function describing the dynamics of the system, since `model` and `data` were required arguments for the integration anyway. 

----
📚 Documentation preview 📚: https://jaxsim--252.org.readthedocs.build//252/


flferretti:
@diegoferigo I modified the logic to be backward compatible, so it should still be compatible with parametric models

src/jaxsim/api/model.py
```diff
@@ -210,15 +234,44 @@ def build(
...
+                    dynamics=js.ode.wrap_system_dynamics_for_integration(
+                        system_dynamics=js.ode.system_dynamics
+                    )
+                )
+
+            # If an integrator class is passed, assume that it is a JaxSim integrator
+            # and build it with the default system dynamics.
+            case _ if isinstance(
+                jaxsim.integrators.Integrator, type(jaxsim.integrators.Integrator)
+            ):
```
flferretti, line 253: This will always evaluate to True, skipping the default case. What about this:
```suggestion
            case _ if issubclass(
                integrator, jaxsim.integrators.Integrator
            ):
```

src/jaxsim/api/model.py
```diff
@@ -2056,7 +2111,7 @@ def step(
             jaxsim.exceptions.raise_runtime_error_if(
                 condition=jnp.logical_and(
                     isinstance(
-                        integrator,
+                        model.integrator,
```
flferretti, line 2114: My bad, we should leave `integrator` here as it is the one used for stepping the simulation 
```suggestion
                        integrator,
```

flferretti (commented):
Early comments

src/jaxsim/api/model.py
diegoferigo: Done in https://github.com/ami-iit/jaxsim/pull/252/commits/68410809a1454a7225d52d8f4e09940ff69c7e13

src/jaxsim/api/model.py
diegoferigo: Done in https://github.com/ami-iit/jaxsim/pull/252/commits/72346140876228b35c0cfd665367756fbb550f53

diegoferigo (approved):

src/jaxsim/api/model.py
```diff
@@ -54,6 +54,10 @@ class JaxSimModel(JaxsimDataclass):
         default=None, repr=False
     )
 
+    integrator: Static[jaxsim.integrators.Integrator | None] = dataclasses.field(
+        default=None, repr=False
+    )
+
```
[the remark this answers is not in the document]
flferretti: Leaving unresolved

PR #253, opened 2024-10-04
Uniform usage of `joint_forces` and `joint_force_references` argument names

diegoferigo:
In short:

- `joint_forces` are the real joint forces that are set to the joint.
- `joint_force_references` are the desired joint forces, that might differ from the real joint forces in presence of effects like joint friction.

This PR updates the APIs making sure to use the right argument name so that it should be more clear to callers whether they might expect to lose a fraction of the passed joint force due to physical effects.

@flferretti @xela-95 please let me know if I missed something.


xela-95:
> @flferretti @xela-95 please let me know if I missed something.

Thanks @diegoferigo ! This will for sure help to clarify the usage of this argument. 

Just to be sure, have you also checked the use of `joint_force_references` in `jaxsim.api.contact` (like in https://github.com/ami-iit/jaxsim/blob/7a2d193f81b199a14c486dcb226c3ca7acf15703/src/jaxsim/api/contact.py#L128  and https://github.com/ami-iit/jaxsim/blob/7a2d193f81b199a14c486dcb226c3ca7acf15703/src/jaxsim/api/contact.py#L94) and in the `compute_contact_forces` methods of the different contact models (like rigid and relaxed_rigid that use such input arguments)?

xela-95 (approved):

diegoferigo:
> Just to be sure, have you also checked the use of `joint_force_references` in `jaxsim.api.contact`

I already checked them, let me know if you find mistakes :) I have a lot of changes pending, I might have missed something.

diegoferigo:
I just noticed that I missed to commit two calls in the `Rigid` and `RelaxedRigid` contact models. I [force-pushed](https://github.com/ami-iit/jaxsim/compare/b639c822aa07ff006fb6005b2d79a40df387443b..33ad6254d0a137dc173b4a44496c3a212cc6de3f) the changes.

flferretti (approved):
Thanks Diego, LGTM!

PR #256, opened 2024-10-07
Remove `*ContactsState` classes

diegoferigo:
In the past few months, JaxSim has received numerous updates and additions to contact management. We now have three contact models (soon four), and the only one that requires to extend the ODE state vector that is integrated with the floating-base dynamics is the original `SoftContacts` model.

- This PR generalizes how additional variables can be introduced to the integrated state vector.
- The material deformation $\mathbf{m}$ used by the `SoftContacts` model is now treated as an augmented ODE state.
- This generalization allow users to use our integrators with their own custom dynamics (potentially including additional effects not currently considered in JaxSim, e.g. actuator dynamics and muskolo-skeletal models).

----
📚 Documentation preview 📚: https://jaxsim--256.org.readthedocs.build//256/

### DIFF
--- a/src/jaxsim/api/data.py
+++ b/src/jaxsim/api/data.py
@@ -13,7 +13,6 @@ import jaxsim.api as js
 import jaxsim.math
 import jaxsim.rbda
 import jaxsim.typing as jtp
-from jaxsim.rbda.contacts import SoftContacts
 from jaxsim.utils import Mutability
 from jaxsim.utils.tracing import not_tracing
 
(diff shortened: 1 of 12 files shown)


src/jaxsim/api/data.py
```diff
@@ -133,56 +132,73 @@ def build(
...
+            (
+                base_angular_velocity
+                if base_angular_velocity is not None
+                else jnp.zeros(3)
+            ),
+            dtype=float,
         ).squeeze()
 
-        gravity = jnp.zeros(3).at[2].set(-standard_gravity)
+        gravity = jnp.zeros(3, dtype=float).at[2].set(-standard_gravity)
```
flferretti, line 175: The `dtype` of `jnp.zeros` is `float` by default, but we can leave it here for safety
```suggestion
        gravity = jnp.zeros(3).at[2].set(-standard_gravity)
```

flferretti (changes requested):
Thank you Diego! This is great. I just left a single minor comment

src/jaxsim/api/data.py
diegoferigo: Can be removed, done in https://github.com/ami-iit/jaxsim/pull/256/commits/247826f66dee036f4779729092902408cf644ba1.

src/jaxsim/rbda/contacts/rigid.py
```diff
@@ -113,6 +91,34 @@ class RigidContacts(ContactModel):
         default_factory=FlatTerrain
```
xela-95, line 91: Just to be coherent with the other contact models:

```suggestion
        default_factory=FlatTerrain.build
```

xela-95 (approved):
Thanks for the improvement @diegoferigo !! I just left a minor comment

src/jaxsim/api/data.py
diegoferigo: @flferretti not sure if you realized that you blocked merging by asking changes. Given the nature of your comment, maybe it was a mistake :)

flferretti (approved):

PR #260, opened 2024-10-09
Refactor contact models and add tests

diegoferigo:
I wanted to polish and uniform the great work done in the last few months about contact models. Likely, we need to keep working on them to optimize their performance, I took advantage of going through them more carefully than my previous reviews.

I have some doubts on few details of the `RigidContacts` and `RelaxedRigidContacts`. I will comment my edits in this PR to discuss my doubts and proposed changes.

----
📚 Documentation preview 📚: https://jaxsim--260.org.readthedocs.build//260/

### DIFF
--- a/src/jaxsim/api/contact.py
+++ b/src/jaxsim/api/contact.py
@@ -9,7 +9,6 @@ import jaxsim.api as js
 import jaxsim.exceptions
 import jaxsim.terrain
 import jaxsim.typing as jtp
-from jaxsim import logging
 from jaxsim.math import Adjoint, Cross, Transform
 from jaxsim.rbda import contacts
 
@@ -303,6 +302,58 @@ def in_contact(
     return links_in_contact
 
 
+def estimate_good_contact_parameters(
+    model: js.model.JaxSimModel,
+    *,
+    standard_gravity: jtp.FloatLike = jaxsim.math.StandardGravity,
+    static_friction_coefficient: jtp.FloatLike = 0.5,
+    number_of_active_collidable_points_steady_state: jtp.IntLike = 1,
+    damping_ratio: jtp.FloatLike = 1.0,
+    max_penetration: jtp.FloatLike | None = None,
+    **kwargs,
+) -> jaxsim.rbda.contacts.ContactParamsTypes:
+    """
+    Estimate good contact parameters.
+
+    Args:
+        model: The model to consider.
+        standard_gravity: The standard gravity constant.
+        static_friction_coefficient: The static friction coefficient.
+        number_of_active_collidable_points_steady_state:
+            The number of active collidable points in steady state supporting
+            the weight of the robot.
+        damping_ratio: The damping ratio.
+        max_penetration:
+            The maximum penetration allowed in steady state when the robot is
+            supported by the configured number of active collidable points.
+        kwargs:
+            Additional model-specific parameters passed to the builder method of
+            the parameters class.
+
+    Returns:
+        The estimated good soft contacts parameters.
+
+    Note:
+        This is primarily a convenience function for soft-like contact models.
+        However, it provides with some good default parameters also for the other ones.
+
+    Note:
+        This method provides a good set of contacts parameters.
+        The user is encouraged to fine-tune the parameters based on the
+        specific application.
+    """
+
+    return estimate_good_soft_contacts_parameters(
+        model=model,
+        standard_gravity=standard_gravity,
+        static_friction_coefficient=static_friction_coefficient,
+        number_of_active_collidable_points_steady_state=number_of_active_collidable_points_steady_state,
+        damping_ratio=damping_ratio,
+        max_penetration=max_penetration,
+        **kwargs,
+    )
+
+
 def estimate_good_soft_contacts_parameters(
     model: js.model.JaxSimModel,
     *,
@@ -359,6 +410,7 @@ def estimate_good_soft_contacts_parameters(
     max_δ = (
         max_penetration
         if max_penetration is not None
+        # Consider as default a 0.5% of the model height.
         else 0.005 * estimate_model_height(model=model)
     )
 
@@ -376,8 +428,11 @@ def estimate_good_soft_contacts_parameters(
                 max_penetration=max_δ,
                 number_of_active_collidable_points_steady_state=nc,
                 damping_ratio=damping_ratio,
-                p=model.contact_model.parameters.p,
-                q=model.contact_model.parameters.q,
+                **dict(
+                    p=model.contact_model.parameters.p,
+                    q=model.contact_model.parameters.q,
+                )
+                | kwargs,
             )
 
         case contacts.ViscoElasticContacts():
@@ -391,15 +446,40 @@ def estimate_good_soft_contacts_parameters(
                     max_penetration=max_δ,
                     number_of_active_collidable_points_steady_state=nc,
                     damping_ratio=damping_ratio,
-                    p=model.contact_model.parameters.p,
-                    q=model.contact_model.parameters.q,
-                    **kwargs,
+                    **dict(
+                        p=model.contact_model.parameters.p,
+                        q=model.contact_model.parameters.q,
+                    )
+                    | kwargs,
                 )
             )
 
+        case contacts.RigidContacts():
+            assert isinstance(model.contact_model, contacts.RigidContacts)
+
+            # Disable Baumgarte stabilization by default since it does not play
+            # well with the forward Euler integrator.
+            K = kwargs.get("K", 0.0)
+
+            parameters = contacts.RigidContactsParams.build(
+                mu=static_friction_coefficient,
+                **dict(
+                    K=K,
+                    D=2 * jnp.sqrt(K),
+                )
+                | kwargs,
+            )
+
+        case contacts.RelaxedRigidContacts():
+            assert isinstance(model.contact_model, contacts.RelaxedRigidContacts)
+
+            parameters = contacts.RelaxedRigidContactsParams.build(
+                mu=static_friction_coefficient,
+                **kwargs,
+            )
+
         case _:
-            logging.warning("The active contact model is not soft-like, no-op.")
-            parameters = model.contact_model.parameters
+            raise ValueError(f"Invalid contact model: {model.contact_model}")
 
     return parameters
 
--- a/src/jaxsim/api/contact.py
+++ b/src/jaxsim/api/contact.py
@@ -35,11 +35,10 @@ def collidable_point_kinematics(
         the linear component of the mixed 6D frame velocity.
     """
 
-    from jaxsim.rbda import collidable_points
-
     # Switch to inertial-fixed since the RBDAs expect velocities in this representation.
     with data.switch_velocity_representation(VelRepr.Inertial):
-        W_p_Ci, W_ṗ_Ci = collidable_points.collidable_points_pos_vel(
+
+        W_p_Ci, W_ṗ_Ci = jaxsim.rbda.collidable_points.collidable_points_pos_vel(
             model=model,
             base_position=data.base_position(),
             base_quaternion=data.base_orientation(dcm=False),
--- a/src/jaxsim/api/contact.py
+++ b/src/jaxsim/api/contact.py
@@ -312,12 +312,7 @@ def estimate_good_soft_contacts_parameters(
     damping_ratio: jtp.FloatLike = 1.0,
     max_penetration: jtp.FloatLike | None = None,
     **kwargs,
-) -> (
-    jaxsim.rbda.contacts.RelaxedRigidContactsParams
-    | jaxsim.rbda.contacts.RigidContactsParams
-    | jaxsim.rbda.contacts.SoftContactsParams
-    | jaxsim.rbda.contacts.ViscoElasticContactsParams
-):
+) -> jaxsim.rbda.contacts.ContactParamsTypes:
     """
     Estimate good parameters for soft-like contact models.
 
--- a/src/jaxsim/api/contact.py
+++ b/src/jaxsim/api/contact.py
@@ -9,6 +9,7 @@ import jaxsim.api as js
 import jaxsim.exceptions
 import jaxsim.terrain
 import jaxsim.typing as jtp
+from jaxsim import logging
 from jaxsim.math import Adjoint, Cross, Transform
 from jaxsim.rbda import contacts
 
@@ -301,6 +302,15 @@ def in_contact(
     return links_in_contact
 
 
+def estimate_good_soft_contacts_parameters(
+    *args, **kwargs
+) -> jaxsim.rbda.contacts.ContactParamsTypes:
+
+    msg = "This method is deprecated, please use `{}`."
+    logging.warning(msg.format(estimate_good_contact_parameters.__name__))
+    return estimate_good_contact_parameters(*args, **kwargs)
+
+
 def estimate_good_contact_parameters(
     model: js.model.JaxSimModel,
     *,
@@ -330,7 +340,7 @@ def estimate_good_contact_parameters(
             the parameters class.
 
     Returns:
-        The estimated good soft contacts parameters.
+        The estimated good contacts parameters.
 
     Note:
         This is primarily a convenience function for soft-like contact models.
@@ -338,51 +348,6 @@ def estimate_good_contact_parameters(
 
     Note:
         This method provides a good set of contacts parameters.
-        The user is encouraged to fine-tune the parameters based on the
-        specific application.
-    """
-
-    return estimate_good_soft_contacts_parameters(
-        model=model,
-        standard_gravity=standard_gravity,
-        static_friction_coefficient=static_friction_coefficient,
-        number_of_active_collidable_points_steady_state=number_of_active_collidable_points_steady_state,
-        damping_ratio=damping_ratio,
-        max_penetration=max_penetration,
-        **kwargs,
-    )
-
-
-def estimate_good_soft_contacts_parameters(
-    model: js.model.JaxSimModel,
-    *,
-    standard_gravity: jtp.FloatLike = jaxsim.math.StandardGravity,
-    static_friction_coefficient: jtp.FloatLike = 0.5,
-    number_of_active_collidable_points_steady_state: jtp.IntLike = 1,
-    damping_ratio: jtp.FloatLike = 1.0,
-    max_penetration: jtp.FloatLike | None = None,
-    **kwargs,
-) -> jaxsim.rbda.contacts.ContactParamsTypes:
-    """
-    Estimate good parameters for soft-like contact models.
-
-    Args:
-        model: The model to consider.
-        standard_gravity: The standard gravity constant.
-        static_friction_coefficient: The static friction coefficient.
-        number_of_active_collidable_points_steady_state:
-            The number of active collidable points in steady state supporting
-            the weight of the robot.
-        damping_ratio: The damping ratio.
-        max_penetration:
-            The maximum penetration allowed in steady state when the robot is
-            supported by the configured number of active collidable points.
-
-    Returns:
-        The estimated good soft contacts parameters.
-
-    Note:
-        This method provides a good starting point for the soft contacts parameters.
         The user is encouraged to fine-tune the parameters based on the
         specific application.
     """
--- a/src/jaxsim/rbda/contacts/common.py
+++ b/src/jaxsim/rbda/contacts/common.py
@@ -1,7 +1,11 @@
 from __future__ import annotations
 
 import abc
+import functools
 from typing import Any
+
+import jax
+import jax.numpy as jnp
 
 import jaxsim.api as js
 import jaxsim.terrain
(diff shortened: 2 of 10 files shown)


src/jaxsim/rbda/contacts/rigid.py
```diff
@@ -485,45 +458,48 @@ def _linear_acceleration_of_collidable_points(
...
-                    h,
-                    ḣ,
-                    k_baumgarte,
-                    d_baumgarte,
-                ),
+                lambda δ, δ̇̇, K, D: jnp.zeros(3),
+                # This is equivalent to: K*(pT - p)⋅n̂ + D*(0 - v)⋅n̂,
+                # where pT is the point on the terrain surface vertical to p.
+                lambda δ, δ̇, K, D: jnp.zeros(3).at[2].set(K * δ + D * δ̇),
+                *(δ, δ̇, k_baumgarte, d_baumgarte),
```
diegoferigo, line 496: As discussed f2f, I have the impression that here there was an error in the sign of the $D \dot{\mathbf{h}}$ term.

I refactored the code to use the normal material deformation $\delta$ instead of the penetration depth $\mathbf{h}$, that are related as follows:

```math
\begin{cases}
\delta &= \mathbf{h} \cdot \hat{\boldsymbol{n}} = -(\mathcal{H}(\mathbf{p}) - \mathbf{p}) \cdot \hat{\boldsymbol{n}} \\
\dot{\delta} &= (\dot{\mathbf{p}} \cdot \hat{\boldsymbol{n}}) \: \hat{\boldsymbol{n}}
\end{cases}
```

src/jaxsim/rbda/contacts/relaxed_rigid.py
```diff
@@ -271,34 +273,33 @@ def compute_contact_forces(
...
-        δ = jax.vmap(_detect_contact)(*position.T)
+        # Compute the penetration depth and velocity of the collidable points.
+        # Note that this function considers the penetration in the normal direction.
+        δ, δ̇, n̂ = jax.vmap(common.compute_penetration_data, in_axes=(0, 0, None))(
+            position, velocity, model.terrain
+        )
+
+        # Compute the position and velocity in the constraint frame.
+        position_constraint = jax.vmap(lambda δ, n̂: -δ * n̂)(δ, n̂)
+        velocity_constraint = jax.vmap(lambda δ̇, n̂: -δ̇ * n̂)(δ̇, n̂)
```
diegoferigo, line 290: Note that this contact model calls _constraint frame_ what we call $C[W]$, at least to my understanding. cc @flferretti

src/jaxsim/rbda/contacts/relaxed_rigid.py
```diff
@@ -360,25 +386,35 @@ def step(carry):
...
+        # ======================================
+        # Compute the contact forces with L-BFGS
+        # ======================================
+
+        # Initialize the optimized forces with a linear Hunt/Crossley model.
         init_params = (
-            K[:, jnp.newaxis] * jnp.zeros_like(position).at[:, 2].set(δ)
-            + D[:, jnp.newaxis] * velocity
+            K[:, jnp.newaxis] * position_constraint
+            + D[:, jnp.newaxis] * velocity_constraint
```
diegoferigo, line 417: @flferretti this warm start is now computed with the full normal component computed with Hunt/Crossley model. Before, only the z component was used, that was true only on flat terrain.

src/jaxsim/rbda/contacts/relaxed_rigid.py
```diff
@@ -390,124 +426,153 @@ def continuing_criterion(carry):
...
+                vel=vel,
             )
 
-            # Compute the regularization terms.
+            # Compute the regularization term.
             R = (
                 (2 * μ**2 * (1 - ξ) / (ξ + 1e-12))
                 * (1 + μ**2)
-                @ jnp.linalg.inv(M_L[link_idx, :3, :3])
+                * (1 / jnp.diag(M_L[link_idx, 0:3, 0:3]).mean())
```
diegoferigo, line 556: @flferretti I suspect that the usage of the spatial inertia of the link $\mathbb{M}\_L$ is wrong here. Accordingly to the following resources, it seems to me that Mujoco initializes it to $\text{diag}(J^\top_{W,C} M^{-1} J_{W,C})[0:3, 0:3]$ computed at $\mathbf{q}_0$ (default configuration from the URDF). It's not a big deal now, let's keep in mind to validate it as soon as possible.

- https://github.com/google/brax/blob/865d974c0285d345425d76df30c38614c828c8c8/brax/generalized/constraint.py
- https://github.com/google/brax/blob/865d974c0285d345425d76df30c38614c828c8c8/brax/base.py#L316C5-L316C14
- https://github.com/google-deepmind/mujoco/blob/5e989d18c349e914fb82ae937006ff4832b011ce/src/engine/engine_setconst.c#L87-L100

diegoferigo:
In this PR, I also added some simple test of all contact models so that we at least know that we won't break anything related to them in the future. Regardless, since I touched some internal details of the `RigidContacts` and `RelaxedRigidContacts`, please @flferretti @xela-95 during your review make sure that your scripts still work as expected.

flferretti (commented):
Just an early comment, would it make sense to include in this PR avoid repeating the `contact_params` attributes in `JaxSimModelData` and in `ContactModel`?

src/jaxsim/rbda/contacts/relaxed_rigid.py
flferretti: Thanks Diego, you're right! This might be substituted with a different warm-start anyway

src/jaxsim/rbda/contacts/rigid.py
xela-95: Nice catch thanks! It seems sounds to me

diegoferigo:
@xela-95 w.r.t. to the original PR, I updated the direction of action of the baumgarte stabilization. Now it produces a force towards the terrain's normal, and this behavior is aligned with `RelaxedRigidContacts`. While reviewing, do you mind running the walking script with Baumgarte stabilization enabled? I'm curious to see if it works well now.

> Just an early comment, would it make sense to include in this PR avoid repeating the `contact_params` attributes in `JaxSimModelData` and in `ContactModel`?

Let me think about it. Having nominal parameters already defined in the contact model itself is quite useful, but the dynamic override from `data` is not super intuitive. I'm in favor of this change, but I need to play a bit with it before proceeding. I'll try to do it in a new PR.

src/jaxsim/rbda/contacts/relaxed_rigid.py
```diff
@@ -360,25 +386,35 @@ def step(carry):
...
                 continuing_criterion, step, init_carry
             )
+
             return final_params, final_state
 
+        # ======================================
+        # Compute the contact forces with L-BFGS
+        # ======================================
+
+        # Initialize the optimized forces with a linear Hunt/Crossley model.
```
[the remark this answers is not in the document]
diegoferigo: Considering the no-side-effects JAX pattern, I'm a bit hesitant on this type of warm start since it would introduce an extra state. While using the last solution is the optimal warm start, before introducing the complexity of this extra state, we should understand if using a simpler method like the existing one provides a comparable speed-up.

src/jaxsim/rbda/contacts/common.py
```diff
@@ -14,6 +18,47 @@
...
+    # Compute the penetration depth normal to the terrain.
+    δ = jnp.maximum(0.0, jnp.dot(h, n̂))
+
+    # Compute the penetration normal velocity.
+    δ̇ = -jnp.dot(W_ṗ_C, n̂)
+
+    # Enforce the penetration rate to be zero when the penetration depth is zero.
+    δ̇ = jnp.where(δ > 0, δ̇, 0.0)
+
+    return δ, δ̇, n̂
```
traversaro, line 59: I guess this is a Unicode problem of the accent not handled correctly, but this is a bit confusing, as the dot goes in the subsequent comma at least on my machine:

unicodeaccent

src/jaxsim/rbda/contacts/rigid.py
```diff
@@ -372,64 +333,74 @@ def compute_contact_forces(
...
+        # Compute stabilization term.
         baumgarte_term = RigidContacts._compute_baumgarte_stabilization_term(
-            inactive_collidable_points=inactive_collidable_points,
-            h=h,
-            ḣ=ḣ,
-            K=self.parameters.K,
-            D=self.parameters.D,
+            inactive_collidable_points=(δ <= 0),
+            δ=δ,
+            δ̇=δ̇,
```
traversaro, line 349: See https://github.com/ami-iit/jaxsim/pull/260/files#r1795057708, this is even more confusing:

~~~
δ̇=δ̇,
~~~

src/jaxsim/rbda/contacts/common.py
traversaro: We checked with @xela-95, and the problem is not there in Chrome, but it is in Visual Studio Code:

image

perhaps we can try to avoid to use this character until it is supported in Visual Studio Code? I guess not a lot of people use Firefox, but Visual Studio Code instead is quite used.

diegoferigo: Generally speaking, I usually never use unicode characters in public APIs like function arguments. This would mean that if the user wants to pass in their downstream code a `kwarg`, they have to know how to deal with unicode. This is something to avoid.

The situation is different for logic in function body. Here, I value more having code that is 1) readable and 2) as close as possible to pseudocode. Dotted variables are use pretty much everywhere in JaxSim, I'm curious, why this case is different than all other ones?

traversaro: I totally like dotted variables (when they work). My guess is that the difference here is that diacritics with greek letters are not supported uniformly across fonts.

src/jaxsim/api/contact.py
```diff
@@ -303,6 +301,58 @@ def in_contact(
...
+
+    return estimate_good_soft_contacts_parameters(
+        model=model,
+        standard_gravity=standard_gravity,
+        static_friction_coefficient=static_friction_coefficient,
+        number_of_active_collidable_points_steady_state=number_of_active_collidable_points_steady_state,
+        damping_ratio=damping_ratio,
+        max_penetration=max_penetration,
+        **kwargs,
+    )
```
flferretti, line 353: Does this method differs in any way from `estimate_good_soft_contacts_parameters`? If I understood well, this should replace the aforementioned method in future releases, so probably we could start a deprecation cycle in which we can rename `estimate_good_soft_contacts_parameters` to `_estimate_good_soft_contacts_parameters`, and then implement `estimate_good_soft_contacts_parameters` as:

```python
def estimate_good_soft_contacts_parameters(*args, **kwargs):
    logging.warning("This method is deprecated, please use `estimate_good_contact_parameters`)
    return estimate_good_contacts_parameters(*args, **kwargs)
```

tests/test_simulations.py
```diff
@@ -192,3 +194,237 @@ def test_box_with_zero_gravity(
...
+    # Check jaxsim_model_box@conftest.py.
+    box_height = 0.1
+
+    # Build the data of the model.
+    data_t0 = js.data.JaxSimModelData.build(
+        model=model,
+        base_position=jnp.array([0.0, 0.0, box_height * 2]),
+        velocity_representation=VelRepr.Inertial,
+        # In order to achieve almost no penetration, we need to use a fairly large
+        # Baumgarte stabilization term.
```
flferretti, line 412: Maybe it was related to a different contact model
```suggestion
```

tests/test_simulations.py
```diff
@@ -192,3 +194,237 @@ def test_box_with_zero_gravity(
...
+        contacts_params=js.contact.estimate_good_contacts_parameters(
+            model=model,
+            time_constant=0.001,
+        ),
```
flferretti, line 416: Is `js.contact.estimate_good_contacts_parameters` needed here?

src/jaxsim/rbda/contacts/relaxed_rigid.py
```diff
@@ -390,124 +425,153 @@ def continuing_criterion(carry):
...
+        ) -> tuple[jtp.Vector, jtp.Matrix, jtp.Float, jtp.Float]:
 
             # Compute the reference acceleration.
-            ξ, a_ref, K, D = _imp_aref(
-                penetration=penetration,
-                velocity=velocity,
+            ξ, a_ref, K, D = imp_aref(
+                pos=pos,
+                vel=vel,
             )
```
flferretti, line 549: Probably by removing a comma we can put this in a single line
```suggestion
            ξ, a_ref, K, D = imp_aref(pos=pos, vel=vel)
```

flferretti (changes requested):

src/jaxsim/rbda/contacts/relaxed_rigid.py
diegoferigo: Yep, done in https://github.com/ami-iit/jaxsim/pull/260/commits/85a98dabe7afa88ef9b7c913d4b8875e2346e244.

tests/test_simulations.py
diegoferigo: Thanks! Removed in https://github.com/ami-iit/jaxsim/pull/260/commits/85a98dabe7afa88ef9b7c913d4b8875e2346e244.

src/jaxsim/api/contact.py
diegoferigo: Done in https://github.com/ami-iit/jaxsim/pull/260/commits/8c884258469359bd4303151aa4faeba5262d847e.

tests/test_simulations.py
diegoferigo: Nope, I was playing with the parameters in order to keep the default pytest tolerances, but I was not successful. With the current tolerances, the default parameters are good enough.

src/jaxsim/rbda/contacts/common.py
diegoferigo: Ok now I understood your point. I agree that in few cases, dotted greek letters look strange. Addressed in https://github.com/ami-iit/jaxsim/compare/2b6f69d4755d37dbf5972c074d6eda5f8951ecb1..85a98dabe7afa88ef9b7c913d4b8875e2346e244, thanks!

src/jaxsim/rbda/contacts/rigid.py
diegoferigo: Same as https://github.com/ami-iit/jaxsim/pull/260#discussion_r1795722266.

src/jaxsim/rbda/contacts/common.py
traversaro: Great, thanks! Perhaps in some time the default fonts in Visual Studio Code will address the problem and we can start using it again.

diegoferigo: For the records, my Python IDE is PyCharm with the **Fira Code** font (that supports ligatures, which are beautiful), and dotted greek letters look as follows:

image

traversaro: Yes, also on Google Chrome it works fine, it is probably an issue of default fonts.

src/jaxsim/rbda/contacts/relaxed_rigid.py
```diff
@@ -243,25 +243,26 @@ def compute_contact_forces(
...
         # Initialize the model and data this contact model is operating on.
         # This will raise an exception if either the contact model or the
         # contact parameters are not compatible.
         model, data = self.initialize_model_and_data(model=model, data=data)
+        assert isinstance(data.contacts_params, RelaxedRigidContactsParams)
 
-        link_forces = (
-            link_forces
+        link_forces = jnp.atleast_2d(
+            jnp.array(link_forces, dtype=float).squeeze()
```
flferretti, line 258: Maybe this is a bit redundant as we are calling `.astype(float)` in line 259
```suggestion
            jnp.array(link_forces).squeeze()
```

src/jaxsim/rbda/contacts/relaxed_rigid.py
```diff
@@ -243,25 +243,26 @@ def compute_contact_forces(
...
             if link_forces is not None
             else jnp.zeros((model.number_of_links(), 6))
-        )
+        ).astype(float)
 
-        joint_force_references = (
-            joint_force_references
+        joint_force_references = jnp.atleast_1d(
+            jnp.array(joint_force_references, dtype=float).squeeze()
```
flferretti, line 264: Same as above, probably a bit redundant 
```suggestion
            jnp.array(joint_force_references).squeeze()
```

src/jaxsim/rbda/contacts/relaxed_rigid.py
```diff
@@ -271,10 +272,10 @@ def compute_contact_forces(
             joint_force_references=joint_force_references,
         )
 
-        def _detect_contact(x: jtp.Array, y: jtp.Array, z: jtp.Array) -> jtp.Array:
+        def detect_contact(x: jtp.Array, y: jtp.Array, z: jtp.Array) -> jtp.Array:
             x, y, z = jax.tree.map(jnp.squeeze, (x, y, z))
 
-            n̂ = self.terrain.normal(x=x, y=y).squeeze()
+            n̂ = model.terrain.normal(x=x, y=y).squeeze()
```
flferretti, line 280: Do we still need `terrain` as a class attribute? Can it just be extracted from `model` ?

src/jaxsim/rbda/contacts/relaxed_rigid.py
```diff
@@ -286,19 +287,19 @@ def _detect_contact(x: jtp.Array, y: jtp.Array, z: jtp.Array) -> jtp.Array:
         )
 
         # Compute the activation state of the collidable points
-        δ = jax.vmap(_detect_contact)(*position.T)
+        δ = jax.vmap(detect_contact)(*position.T)
```
flferretti, line 292: Probably we can use `compute_penetration_data` directly

```suggestion
        δ, δ_dot, _ = contacts.common.compute_penetration_data(p=position, v=velocity, terrain=model.terrain)
```

src/jaxsim/rbda/contacts/relaxed_rigid.py
```diff
@@ -360,22 +386,32 @@ def step(carry):
...
             return final_params, final_state
 
+        # ======================================
+        # Compute the contact forces with L-BFGS
+        # ======================================
+
+        # Initialize the optimized forces with a linear Hunt/Crossley model.
         init_params = (
             K[:, jnp.newaxis] * jnp.zeros_like(position).at[:, 2].set(δ)
             + D[:, jnp.newaxis] * velocity
```
flferretti, line 419: At this point we can use the normal penetration velocity
```suggestion
            + D[:, jnp.newaxis] * jnp.zeros_like(position).at[:, 2].set(δ_dot)
```

src/jaxsim/rbda/contacts/relaxed_rigid.py
```diff
@@ -390,28 +426,30 @@ def continuing_criterion(carry):
         maxiter = solver_options.pop("maxiter")
 
         # Compute the 3D linear force in C[W] frame.
-        CW_f_Ci, _ = run_optimization(
+        solution, _ = run_optimization(
             init_params=init_params,
-            A=A,
-            b=b,
```
flferretti, line 396: The two kwargs `A` and `b` were placed here for enabling computing the gradient, but if it's not a priority, we can remove them for the time being

src/jaxsim/rbda/contacts/rigid.py
```diff
@@ -295,60 +256,58 @@ def compute_contact_forces(
...
+        assert isinstance(data.contacts_params, RigidContactsParams)
 
-        # Import qpax just in this method
+        # Import qpax privately just in this method.
         import qpax
 
-        link_forces = (
-            link_forces
+        link_forces = jnp.atleast_2d(
+            jnp.array(link_forces, dtype=float).squeeze()
```
flferretti, line 272: Same as above, probably redundant
```suggestion
            jnp.array(link_forces).squeeze()
```

src/jaxsim/rbda/contacts/rigid.py
```diff
@@ -295,60 +256,58 @@ def compute_contact_forces(
...
             if link_forces is not None
             else jnp.zeros((model.number_of_links(), 6))
-        )
+        ).astype(float)
 
-        joint_force_references = (
-            joint_force_references
+        joint_force_references = jnp.atleast_1d(
+            jnp.array(joint_force_references, dtype=float).squeeze()
```
flferretti, line 278: Same as above, probably redundant

```suggestion
            jnp.array(joint_force_references).squeeze()
```

flferretti (changes requested):
Thanks for working on this! I left a couple comments

src/jaxsim/rbda/contacts/relaxed_rigid.py
diegoferigo: The two functions are not compatible. It's part of the refactoring that will be done in the next PR.

src/jaxsim/rbda/contacts/relaxed_rigid.py
diegoferigo: Same as above. I noticed that this contact model is very sensitive to this warm start. It's something worth investigating.

src/jaxsim/rbda/contacts/relaxed_rigid.py
diegoferigo: Since it's part of the `ContactModel` base class, this is not the right place to remove it. As for `ContactModel.parameters` that are overriden by those in `data`.

I still have to wrap my head around those stored in the contact model itsels, as right now they act as default ones.

src/jaxsim/rbda/contacts/rigid.py
diegoferigo: Done in https://github.com/ami-iit/jaxsim/compare/7b2fc3bfe40a8b4998b768df2a2b0f5d959a8e2a..f2e613f4fafe2288f0e4114ed64dbced4a160f39

src/jaxsim/rbda/contacts/rigid.py
diegoferigo: Done in https://github.com/ami-iit/jaxsim/compare/7b2fc3bfe40a8b4998b768df2a2b0f5d959a8e2a..f2e613f4fafe2288f0e4114ed64dbced4a160f39

src/jaxsim/rbda/contacts/relaxed_rigid.py
diegoferigo: Yep let's keep this in mind as soon as this contact model starts working reliably. There are many things I'm investigating that do not sum up (friction cone not met, possibility to apply bilateral forces, etc).

diegoferigo:
@flferretti @xela-95 I removed from this PR the new logic of the `RelaxedRigid` contact model. I need some more time to investigate how to properly update it, as there are behaviors that do not sum up. The `comodo` notebook we were discussing f2f should work with these changes.

There are few PRs that depend on this one. I want to merge these changes and propose again the new logic of the `RelaxedRigid` contact model in a new one.

src/jaxsim/rbda/contacts/relaxed_rigid.py
diegoferigo: Delayed to a next PR.

PR #262, opened 2024-10-11
Remove absolute time from `JaxSimModelData` and introduce `JaxSimModel.time_step`

diegoferigo:
In this PR, I'm finally addressing a longstanding problem that complicated the usage of 32-bit precision as described in #136. After having explored more complex solutions, we decided to stop tracking the absolute simulation time within JaxSim, moving the responsibility to downstream code if necessary.

More details:

- Remove `JaxSimData.time()` and `JaxSimData.time_ns`.
- Add a new `JaxSimModel.time_step` that stores the default integration $\Delta t$. This can be useful in some particular cases, for example those system dynamics that need to know the $\Delta t$ used by the integrator (e.g. semi-implicit integrators, visco-elastic contacts, etc).
- Add a new `t0` argument to `jaxsim.api.model.step`, so that it's still possible to run AD w.r.t. the time if the dynamics considered by the integrator is time-dependent (this is rarely the case). 

----
📚 Documentation preview 📚: https://jaxsim--262.org.readthedocs.build//262/

### DIFF
--- a/src/jaxsim/api/ode.py
+++ b/src/jaxsim/api/ode.py
@@ -63,7 +63,6 @@ def wrap_system_dynamics_for_integration(
         # Update the state and time stored inside data.
         with data_f.editable(validate=True) as data_rw:
             data_rw.state = x
-            data_rw.time_ns = jnp.array(t * 1e9).astype(data_rw.time_ns.dtype)
 
         # Evaluate the system dynamics, allowing to override the kwargs originally
         # passed when the closure was created.
--- a/src/jaxsim/rbda/contacts/visco_elastic.py
+++ b/src/jaxsim/rbda/contacts/visco_elastic.py
@@ -928,7 +928,6 @@ class ViscoElasticContacts(common.ContactModel):
         # Create the data at the final time.
         with data.editable(validate=True) as data_tf:
             data_tf: js.data.JaxSimModelData
-            data_tf.time_ns = data.time_ns + (dt * 1e9).astype(data.time_ns.dtype)
 
         data_tf = data_tf.reset_joint_positions(q_plus[7:])
         data_tf = data_tf.reset_base_position(q_plus[0:3])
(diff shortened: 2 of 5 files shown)


src/jaxsim/rbda/contacts/visco_elastic.py
flferretti, line 1: Beware that `data.time_ns` is still used in line 928

src/jaxsim/api/ode.py
flferretti, line 1: Argument `t` is unused

flferretti (changes requested):
Thanks a lot Diego! This was very much needed! I just left a couple comments

src/jaxsim/rbda/contacts/visco_elastic.py
diegoferigo: I forgot to move that commit to this PR, I have few stacked branches and I missed that. Thanks for noticing!

src/jaxsim/api/ode.py
diegoferigo: Yes this is correct, our system dynamics is not time-dependent, therefore `t` is always ignored. However, integrators operate on a generic `f(x, t)`, therefore the signature must comply with what they expect.

flferretti: Alright, thanks for the explanation!

src/jaxsim/rbda/contacts/visco_elastic.py
```diff
@@ -925,7 +928,6 @@ def integrate_data_with_average_contact_forces(
         # Create the data at the final time.
         with data.editable(validate=True) as data_tf:
             data_tf: js.data.JaxSimModelData
```
flferretti, line 930: Probably this can be removed
```suggestion
```

flferretti (changes requested):

flferretti (approved):

src/jaxsim/rbda/contacts/visco_elastic.py
diegoferigo: Right, done in https://github.com/ami-iit/jaxsim/pull/262/commits/14d84325b489c81c1e38d64d96d0e2ba2252c50f, thanks!

PR #263, opened 2024-10-11
Disable host callbacks and enforce 32bit precision when running on TPU

flferretti:
This PR introduces two main changes to allow runnning simulations on TPUs:
1. host callbacks are disable by default, since they are not fully supported for TPU architectures
2. 32bit precision is enforced when importing `jaxsim`

Example error on `js.model.step`:
```python
[/usr/local/lib/python3.10/dist-packages/jax/_src/interpreters/pxla.py](https://localhost:8080/#) in __call__(self, *args)
   1275           or self.has_host_callbacks):
   1276         input_bufs = self._add_tokens_to_inputs(input_bufs)
-> 1277         results = self.xla_executable.execute_sharded(
   1278             input_bufs, with_tokens=True
   1279         )

XlaRuntimeError: INVALID_ARGUMENT: Unknown send channel 33 for host callbacks
```

----
📚 Documentation preview 📚: https://jaxsim--263.org.readthedocs.build//263/

### DIFF
--- a/src/jaxsim/__init__.py
+++ b/src/jaxsim/__init__.py
@@ -11,34 +11,32 @@ def _jnp_options() -> None:
     # Check if running on TPU
     is_tpu = jax.devices()[0].platform == "tpu"
 
-    # Determine if 64-bit precision is requested
+    # Enable by default 64-bit precision to get accurate physics.
+    # Users can enforce 32-bit precision by setting the following variable to 0.
     use_x64 = os.environ.get("JAX_ENABLE_X64", "1") != "0"
 
-    # Raise an error if 64-bit precision is not allowed on TPU
+    # Notify the user if unsupported 64-bit precision was enforced on TPU.
     if is_tpu and use_x64:
         msg = "64-bit precision is not allowed on TPU. Enforcing 32bit precision."
-        logging.error(msg)
+        logging.warning(msg)
+        use_x64 = False
 
-    # Enable 64-bit precision in JAX
-    elif not is_tpu and use_x64:
-        logging.info("Enabling JAX to use 64bit precision")
+    # Enable 64-bit precision in JAX.
+    if use_x64:
+        logging.info("Enabling JAX to use 64-bit precision")
         jax.config.update("jax_enable_x64", True)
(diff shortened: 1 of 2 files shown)


src/jaxsim/__init__.py
```diff
@@ -8,18 +8,34 @@ def _jnp_options() -> None:
...
 
+        # Verify 64-bit precision is correctly set
         if jnp.empty(0, dtype=float).dtype != jnp.empty(0, dtype=np.float64).dtype:
             logging.warning("Failed to enable 64bit precision in JAX")
 
+    # Enforce 32-bit precision on TPU
+    elif is_tpu:
+        logging.warning("JAX is running on TPU; 32bit precision is enforced.")
+
+    # Warn about experimental use of 32-bit precision
```
diegoferigo, line 38:
```suggestion
    # Notify the user if unsupported 64-bit precision was enforced on TPU.
    if is_tpu and use_x64:
        msg = "64-bit precision is not allowed on TPU. Enforcing 32-bit precision."
        logging.warning(msg)
        use_x64 = False

    # Enable 64-bit precision in JAX
    if use_x64:
        logging.info("Enabling JAX to use 64-bit precision")
        jax.config.update("jax_enable_x64", True)

        import jax.numpy as jnp
        import numpy as np

        # Verify that 64-bit precision is correctly set.
        if jnp.empty(0, dtype=float).dtype != jnp.empty(0, dtype=np.float64).dtype:
            logging.warning("Failed to enable 64bit precision in JAX")

    # Warn about experimental use of 32-bit precision.
```

diegoferigo (approved):
Thanks @flferretti! Let me know what do you think of the suggestions below, there were cases that could be handled differently from what you proposed.

Please wait #262 to be merged before this one.

src/jaxsim/__init__.py
flferretti: Thanks Diego! Addressed in https://github.com/ami-iit/jaxsim/pull/263/commits/515522f85c091ec8d2376120d42bb56d0463d6bc

PR #266, opened 2024-10-18
Add dependabot

diegoferigo:
Frankly speaking, dependabot has always annoyed me mainly because it used to open too frequent PRs. However, I've recently discovered that the rate of updates can be configured granularly, and it's also possible to merge together all necessary updates in a single PR.

I'd like to give it a try by setting one update per month to the actions part of our GitHub workflows. We can fine tune it in the future, if needed.

- https://docs.github.com/en/code-security/dependabot
- `github-actions`: https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#github-actions
- `groups`: https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#groups

----
📚 Documentation preview 📚: https://jaxsim--266.org.readthedocs.build//266/

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,11 @@
+# To get started with Dependabot version updates, you'll need to specify which
+# package ecosystems to update and where the package manifests are located.
+# Please see the documentation for all configuration options:
+# https://docs.github.com/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file
+
+version: 2
+updates:
+  - package-ecosystem: "" # See documentation for possible values
+    directory: "/" # Location of package manifests
+    schedule:
+      interval: "weekly"
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -6,6 +6,8 @@ updates:
     directory: /
     schedule:
       interval: monthly
+    # Disable rebasing automatically existing pull requests.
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -0,0 +1,13 @@
...
+  # Check for updates to GitHub Actions every month.
+  - package-ecosystem: github-actions
+    directory: /
+    schedule:
+      interval: monthly
+    # Group updates to a single PR.
+    groups:
+      dependencies:
+        patterns:
+          - '*'
```
flferretti, line 13: This defaults to `auto`, meaning that it would automatically rebase the open PRs if any changer on target branch is detected. Are we sure we want this behavior?
```suggestion
          - '*'
    rebase-strategy: "disabled"      
```

flferretti (approved):
LGTM, thanks! Please check the comment I left as it might be crucial for the intended behavior

.github/dependabot.yml
diegoferigo: Good catch, I think that automatic rebasing is not something that we want, at least for now. Thanks for the heads up! Done in https://github.com/ami-iit/jaxsim/pull/266/commits/144078111dd822ee90f30deb81fc78d4e1224839.

PR #270, opened 2024-10-22
Fix bug in Delassus matrix computation in example notebook

xela-95:


diegoferigo (approved):

flferretti (approved):
Thanks a lot Alessandro, LGTM!

PR #272, opened 2024-10-22
 Finalize minor changes for v0.5 release

flferretti:
@diegoferigo feel free to push any minor updates and changes for the next release.

----
📚 Documentation preview 📚: https://jaxsim--272.org.readthedocs.build//272/

### DIFF
--- a/src/jaxsim/api/link.py
+++ b/src/jaxsim/api/link.py
@@ -4,6 +4,7 @@ from collections.abc import Sequence
 import jax
 import jax.numpy as jnp
 import jax.scipy.linalg
+import numpy as np
 
 import jaxsim.api as js
 import jaxsim.rbda
@@ -96,7 +97,7 @@ def idxs_to_names(
         The names of the links.
     """
 
-    return tuple(idx_to_name(model=model, link_index=idx) for idx in link_indices)
+    return tuple(np.array(model.kin_dyn_parameters.link_names)[list(link_indices)])
 
 
 # =========
--- a/src/jaxsim/api/link.py
+++ b/src/jaxsim/api/link.py
@@ -54,9 +54,7 @@ def idx_to_name(model: js.model.JaxSimModel, *, link_index: jtp.IntLike) -> str:
     """
 
     exceptions.raise_value_error_if(
-        condition=jnp.array(
-            [link_index < 0, link_index >= model.number_of_links()]
-        ).any(),
+        condition=link_index < 0,
         msg="Invalid link index '{idx}'",
         idx=link_index,
     )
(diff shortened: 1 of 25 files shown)


src/jaxsim/api/link.py
```diff
@@ -54,9 +54,7 @@ def idx_to_name(model: js.model.JaxSimModel, *, link_index: jtp.IntLike) -> str:
     """
 
     exceptions.raise_value_error_if(
-        condition=jnp.array(
-            [link_index < 0, link_index >= model.number_of_links()]
-        ).any(),
+        condition=link_index < 0,
```
xela-95, line 57: Why the case `link_index >= model.number_of_links()` has been removed?

xela-95 (approved):
Thanks @flferretti for the cleanup and enhancements! I left just a minor comment.

Approving

src/jaxsim/api/link.py
flferretti: `model.kin_dyn_parameters.link_names` is a `Static[tuple[str, ...]]`, so if the link index is out of bounds, it will raise an `IndexError` anyway

xela-95: Ok thanks!

CarlottaSartore:
@flferretti  @xela-95  recall of this https://github.com/conda-forge/jaxlib-feedstock/issues/285. In case it is not solved before the release let's recall to clip the jaxlib version

src/jaxsim/terrain/terrain.py
```diff
@@ -145,7 +147,7 @@ def height(self, x: jtp.FloatLike, y: jtp.FloatLike) -> jtp.Float:
         D = -C * self._height
 
         # Invert the plane equation to get the height at the given (x, y) coordinates.
-        return jnp.array(-(A * x + B * y + D) / C).astype(float)
+        return jnp.array(-(A * x + B * y + D) / (C + 1e-9)).astype(float)
```
diegoferigo, line 150: Mmh $C=0$ is a vertical plane, right? Is this something that happened?

flferretti: Yes, correct! A possible alternative is to use `jaxsim.exception` to raise a ValueError in case `C = 0`. If you prefer this alternative I can amend the commit

diegoferigo: The small epsilon won't affect that much but if we can avoid it, it would be preferable. Raising an exception is a good option for now.

flferretti: Fixed with rebase in https://github.com/ami-iit/jaxsim/pull/272/commits/99e6ecb3736fd03d6bb772fc3b25d0dfe2c3e979

PR #277, opened 2024-10-30
Fix running AD through `jaxsim.math.Rotation.from_axis_angle`

diegoferigo:
In these days, I tried to solve an unconstrained non-linear optimization problem using the JaxSim functions to construct the cost function. I used the [`L-BFGS`](https://optax.readthedocs.io/en/latest/_collections/examples/lbfgs.html#l-bfgs-solver) solver from optax, and it works flawlessly :tada:

What I'm trying to solve is quite usual in AMI. I want to compute the initial configuration of a humanoid robot such that the knees and the elbows are slightly bent, the soles of the feet are in contact, and the feet are slightly more distant w.r.t. the configuration with zero joint positions. The optimized variables are the joint positions and the base height. The corresponding optimization problem computes FK in the cost function (exploiting AD support through the JaxSim function).

I had to fix the gradient propagation as in this PR to prevent having problems either with zero division in the jitted function  or the same problem in the gradient computed with AD. The trick I used it the _double-where_ workaround, as described in:

- https://jax.readthedocs.io/en/latest/faq.html#gradients-contain-nan-where-using-where
- https://github.com/jax-ml/jax/issues/5039#issuecomment-735430180
- https://github.com/jax-ml/jax/issues/1052#issuecomment-514083352

----
📚 Documentation preview 📚: https://jaxsim--277.org.readthedocs.build//277/

### DIFF
--- a/src/jaxsim/math/rotation.py
+++ b/src/jaxsim/math/rotation.py
@@ -1,4 +1,3 @@
-import jax
 import jax.numpy as jnp
 import jaxlie
 
@@ -8,6 +7,7 @@ from .skew import Skew
 
 
 class Rotation:
+
     @staticmethod
     def x(theta: jtp.Float) -> jtp.Matrix:
         """
@@ -19,6 +19,7 @@ class Rotation:
         Returns:
             jtp.Matrix: 3D rotation matrix.
         """
+
         return jaxlie.SO3.from_x_radians(theta=theta).as_matrix()
 
     @staticmethod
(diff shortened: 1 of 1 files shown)


diegoferigo:
Note: probably a similar trick can be used to remove the small tolerances that we always had in `SoftContacts`:

https://github.com/ami-iit/jaxsim/blob/35cf33692164fe86872a2f3c9e557124353c5013/src/jaxsim/rbda/contacts/soft.py#L288-L292

flferretti (approved):
Thanks Diego, LGTM!

PR #278, opened 2024-10-31
Fix reset of system velocity in `model.step`

xela-95:
This PR fixes a regression on the reset of the system velocity in `model.step` for rigid contacts.

----
📚 Documentation preview 📚: https://jaxsim--278.org.readthedocs.build//278/


flferretti (approved):
Thanks a lot @xela-95! LGTM

diegoferigo (approved):

PR #284, opened 2024-11-13
Raise error if HTTP request fails

flferretti:
This PR modifies an example notebook so that an error is raised when the HTTP request for downloading the cartpole URDF fails.


diegoferigo (approved):

flferretti:
Now we are sure that it is a HTTP error in the RTD action (https://readthedocs.org/projects/jaxsim/builds/26267876/):

```python
[0;31mRuntimeError[0m: Failed to fetch data.
```

PR #294, opened 2024-11-22
Fix simulation test with relaxed rigid contact model

flferretti:
This PR fixes the simulation test with the relaxed rigid contact model by replacing the mass matrix inversion method with `linalg.inv` for accuracy and by using a non-linear Hunt-Crossley model to reduce the drifts in the tangential directions

----
📚 Documentation preview 📚: https://jaxsim--294.org.readthedocs.build//294/

### DIFF
--- a/src/jaxsim/rbda/contacts/relaxed_rigid.py
+++ b/src/jaxsim/rbda/contacts/relaxed_rigid.py
@@ -461,8 +461,8 @@ class RelaxedRigidContacts(common.ContactModel):
                 terrain=model.terrain,
                 K=1e6,
                 D=2e3,
-                p=0.0,
-                q=0.0,
+                p=0.5,
+                q=0.5,
                 # No tangential initial forces.
                 mu=0.0,
                 tangential_deformation=jnp.zeros(3),


src/jaxsim/rbda/contacts/relaxed_rigid.py
```diff
@@ -461,8 +461,8 @@ def continuing_criterion(carry: OptimizationCarry) -> jtp.Bool:
                 terrain=model.terrain,
                 K=1e6,
                 D=2e3,
-                p=0.0,
-                q=0.0,
+                p=0.5,
+                q=0.5,
```
diegoferigo, line 465: This makes sense since we always use a non-linear model with $p$ and $q$ powers you updated. Considering that there's no tangential component ($\mu = 0$) the effect of using different powers is to get an initial guess with a lower magnitude. I don't think it will have a great effect since it's just the starting point, but it's worth investigating.

src/jaxsim/rbda/contacts/relaxed_rigid.py
```diff
@@ -376,7 +376,7 @@ def compute_contact_forces(
 
         # Compute the Delassus matrix and the free mixed linear acceleration of
         # the collidable points.
-        G = Jl_WC @ jnp.linalg.lstsq(M, Jl_WC.T)[0]
+        G = Jl_WC @ jnp.linalg.inv(M) @ Jl_WC.T
```
diegoferigo, line 379: Did you check this change when the free-floating mass matrix $M$ corresponds to almost singular configurations? How comes that by using the plain inverse, the tolerance in the test is now met? I fear that this change might introduce problem on more complex simulations.

xela-95: Yes, I also fear that this could bring some numerical issues in singular configurations.. 
Have you tried to maybe use the `rcond` parameter to see if the issue improves? Or to use [pinv](https://jax.readthedocs.io/en/latest/_autosummary/jax.numpy.linalg.pinv.html)?

src/jaxsim/rbda/contacts/relaxed_rigid.py
flferretti: The test was failing only because of that, so I'm wondering if there is something that can be improved on the optimization part, since it's very sensitive to the initialization

diegoferigo: Yes, this is also what I noticed during the last refactoring.

flferretti:
Thanks @xela-95 and @diegoferigo, I'll change the `lstsq` in a different PR as I believe it is better to restrict this to the scope of #293. Dropping last commit

src/jaxsim/rbda/contacts/relaxed_rigid.py
diegoferigo: I'd go with `pinv` instead of `inv`. The result should be the same as `lstsq`, if it's faster, much better :)

flferretti: Started investigation in https://github.com/ami-iit/jaxsim/pull/295

xela-95 (approved):

PR #295, opened 2024-11-22
Use a `linalg.pinv` to invert the mass matrix in RelaxedRigidContacts

flferretti:
----
📚 Documentation preview 📚: https://jaxsim--295.org.readthedocs.build//295/


xela-95 (approved):
Thanks @flferretti ! LGTM!

PR #299, opened 2024-11-25
Refactor ContactModel to remove `parameters` and `terrain` attributes

### LINKED ISSUE #296: Remove redundant `contacts_params` attribute from `JaxSimModelData`
The `contacts_params` attribute in the `JaxSimModelData` object is redundant, as the same data is accessible directly through `JaxSimModel.contact_model.parameters`. This duplication creates unnecessary complexity and increases maintenance overhead. To improve code simplicity and clarity, I'd propose to remove `contacts_params` from `JaxSimModelData` and utilizing `model.contact_model.parameters` directly within the contact models.

Additionally, this change will eliminate the need to initialize both the `model` and the `data` for each integration step when computing the contact forces. As a result, the `JaxSimModelData` instance will be lighter and more efficient, improving the overall performance of the simulation.

The main disadvantage of this change is that the contact parameters for the SoftContacts model can no longer be initialized automatically. However, the contact parameters attribute should be updated after the JaxSimModel object initialization as follows:

```python
import jaxsim.api as js

model = js.model.JaxSimModel.build_from_model_description(model_description=...)

with model.editable(validate=False) as model:

    model.contact_model = jaxsim.rbda.contacts.SoftContacts.build(
        terrain=model.terrain,
        parameters=js.contact.estimate_good_contact_parameters(
            model=model,
            number_of_active_collidable_points_steady_state=4,
            static_friction_coefficient=1.0,
            damping_ratio=1.0,
            max_penetration=0.001,
        ),
    )
```

Nevertheless, a different way to automatically initialize the `SoftContactsParams` can be investigated.

An initial version of the proposal can be found at [ami-iit/jaxsim@remove_data_contacts_params](https://github.com/ami-iit/jaxsim/tree/remove_data_contacts_params)

flferretti:
This PR removes `parameters` and `terrain` attributes from `ContactModel` as the first will be stored  in `JaxSimModelData` and the latter is always passed as an argument of `compute_contact_forces` and never really used as a class attribute.

Fixes #296


src/jaxsim/rbda/contacts/common.py
```diff
@@ -304,30 +291,17 @@ def zero_state_variables(cls, model: js.model.JaxSimModel) -> dict[str, jtp.Arra
...
         return {}
 
-    def initialize_model_and_data(
-        self,
-        model: js.model.JaxSimModel,
-        data: js.data.JaxSimModelData,
-        validate: bool = True,
-    ) -> tuple[js.model.JaxSimModel, js.data.JaxSimModelData]:
+    @property
+    def _parameter_class(cls) -> type[ContactsParams]:
```
diegoferigo, line 295: Nit

```suggestion
    def _parameters_class(cls) -> type[ContactsParams]:
```

flferretti: Done with rebase in https://github.com/ami-iit/jaxsim/pull/299/commits/da62dffb32453a3c6953fd9ca4f73d1125ec03e9. Thanks!

src/jaxsim/api/model.py
```diff
@@ -42,7 +42,7 @@ class JaxSimModel(JaxsimDataclass):
 
     # Note that this is the default contact model.
     # Its parameters, if any, are then overridden from those stored in JaxSimModelData.
```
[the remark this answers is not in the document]
flferretti: You're right, this should be removed. Addresed in https://github.com/ami-iit/jaxsim/pull/299/commits/b31b62d3a6cffda0d5314fd132c8fa336d44be79

xela-95 (approved):
Thanks @flferretti ! It looks good to me!

PR #300, opened 2024-11-26
Add frame example

CarlottaSartore:
This PR adds  the example for computing velocity and transformation of the frames. In addition, it introduces minor changes to align the frames API with the link's one 

----
📚 Documentation preview 📚: https://jaxsim--300.org.readthedocs.build//300/

### DIFF
--- a/src/jaxsim/api/frame.py
+++ b/src/jaxsim/api/frame.py
@@ -196,7 +196,7 @@ def velocity(
         data: The data of the considered model.
         frame_index: The index of the frame.
         output_vel_repr:
-            The output velocity representation of the link velocity.
+            The output velocity representation of the frame velocity.
 
     Returns:
         The 6D velocity of the frame in the specified velocity representation.


src/jaxsim/api/frame.py
```diff
@@ -180,6 +180,56 @@ def transform(
...
+        idx=frame_index,
+    )
+
+    output_vel_repr = (
+        output_vel_repr if output_vel_repr is not None else data.velocity_representation
+    )
+
+    # Get the link jacobian having I as input representation (taken from data)
+    # and O as output representation, specified by the user (or taken from data).
+    O_J_WL_I = jacobian(
```
xela-95, line 219:
```suggestion
    # Get the frame jacobian having I as input representation (taken from data)
    # and O as output representation, specified by the user (or taken from data).
    O_J_WF_I = jacobian(
```

src/jaxsim/api/frame.py
```diff
@@ -180,6 +180,56 @@ def transform(
...
+        data=data,
+        frame_index=frame_index,
+        output_vel_repr=output_vel_repr,
+    )
+
+    # Get the generalized velocity in the input velocity representation.
+    I_ν = data.generalized_velocity()
+
+    # Compute the link velocity in the output velocity representation.
+    return O_J_WL_I @ I_ν
```
xela-95, line 230:
```suggestion
    # Compute the frame velocity in the output velocity representation.
    return O_J_WF_I @ I_ν
```

xela-95 (changes requested):
Thanks @CarlottaSartore for your contribution! I just left some minor comments. 

Other than this, would you add a little unit test on the `jaxsim.api.frame.velocity` function to avoid regressions in future? It can be added to https://github.com/ami-iit/jaxsim/blob/main/tests/test_api_frame.py

CarlottaSartore:
> Thanks @CarlottaSartore for your contribution! I just left some minor comments.
> 
> Other than this, would you add a little unit test on the `jaxsim.api.frame.velocity` function to avoid regressions in future? It can be added to https://github.com/ami-iit/jaxsim/blob/main/tests/test_api_frame.py

Thanks @xela-95, I have implemented the CI in 5fdbe5b37dc7374ba0bf0600e23fa82abacf13a3

Let me know if there ar any additional changes needed !

CarlottaSartore:
I shoudl have fix with commit 436fde6c3444d9088c09abae5e22876d5eba78b6

The error was that Jaxsim creates a single vector organise as : $[l_0, ..., l_{nl}, f_0,..., f_{nf}]$
What happens is that the idx returned by frame `name_to_idx` is the idx inside such vector and not the index of the vector containing only frames. 
Thus such index should be $>nl$ and $<nl+nf$

xela-95:
> The error was that Jaxsim creates a single vector organise as : [ l 0 , . . . , l n l , f 0 , . . . , f n f ] What happens is that the idx returned by frame `name_to_idx` is the idx inside such vector and not the index of the vector containing only frames. Thus such index should be > n l and < n l + n f

 Right nice catch!

src/jaxsim/api/frame.py
```diff
@@ -180,6 +180,56 @@ def transform(
...
+) -> jtp.Vector:
+    """
+    Compute the 6D velocity of the frame.
+
+    Args:
+        model: The model to consider.
+        data: The data of the considered model.
+        frame_index: The index of the frame.
+        output_vel_repr:
+            The output velocity representation of the link velocity.
```
flferretti, line 199:
```suggestion
            The output velocity representation of the frame velocity.
```

flferretti (changes requested):
Minor typo

xela-95 (approved):
Now it's good for me! :)

CarlottaSartore:
Thanks @xela-95, and @flferretti  I should have also implemented your review,  is it right ? can I proceed with the merge ?

PR #305, opened 2024-12-04
Change neutral element transform matrix in contact

CarlottaSartore:
In this PR we change the neutral element for the transformation matrix from `jnp.zeros` to `jnp.eye` in the `collidable_point_dynamics` method of the contact class. 

This will have no effect on the code execution since such transform matrix was not used, this change is done only to avoid confusion by declaring the correct neutral element for the transformation matrix. 

----
📚 Documentation preview 📚: https://jaxsim--305.org.readthedocs.build//305/

### DIFF
--- a/src/jaxsim/api/contact.py
+++ b/src/jaxsim/api/contact.py
@@ -211,7 +211,7 @@ def collidable_point_dynamics(
     W_H_C = (
         js.contact.transforms(model=model, data=data)
         if data.velocity_representation is not VelRepr.Inertial
-        else jnp.zeros(shape=(len(indices_of_enabled_collidable_points), 4, 4))
+        else jnp.eye(shape=(len(indices_of_enabled_collidable_points), 4, 4))
     )
 
     # Convert the 6D forces to the active representation.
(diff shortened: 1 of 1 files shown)


src/jaxsim/api/contact.py
```diff
@@ -211,7 +211,7 @@ def collidable_point_dynamics(
     W_H_C = (
         js.contact.transforms(model=model, data=data)
         if data.velocity_representation is not VelRepr.Inertial
-        else jnp.zeros(shape=(len(indices_of_enabled_collidable_points), 4, 4))
+        else jnp.eye(shape=(len(indices_of_enabled_collidable_points), 4, 4))
```
flferretti, line 214: `jnp.eye` does not accept `shape`, we should stack them:
```suggestion
        else jnp.stack([jnp.eye(4)] * len(indices_of_enabled_collidable_points))
```

flferretti (changes requested):

src/jaxsim/api/contact.py
CarlottaSartore: You're rigth thanks !

flferretti (approved):
Thanks a lot @CarlottaSartore, LGTM

xela-95 (approved):

PR #307, opened 2024-12-05
Refactor simulation input classes and fix deprecation warnings

flferretti:
This PR removes the `ODEInput` and `PhysicsModelInput` classes to simplify the codebase, and refactors `JaxSimModelReferences` to utilize link forces in inertial representation and joint forces references instead. Moreover, it addresses deprecation warnings for Python 3.16 by updating relevant code. 

----
📚 Documentation preview 📚: https://jaxsim--307.org.readthedocs.build//307/


src/jaxsim/api/references.py
```diff
@@ -26,7 +25,8 @@ class JaxSimModelReferences(js.common.ModelDataWithVelocityRepresentation):
     Class containing the references for a `JaxSimModel` object.
     """
 
-    input: ODEInput
+    _link_forces: jtp.Matrix
```
xela-95, line 28: Maybe we can add a comment to explain what these attributes are and which representation is used to store the link forces?

xela-95 (approved):
Thanks @flferretti! Just minor comments

src/jaxsim/api/references.py
flferretti: Thanks @xela-95! Done in https://github.com/ami-iit/jaxsim/pull/307/commits/baac06cd65443c55ba2eacfc013f63b7e08d546d

PR #319, opened 2024-12-17
Add safe norm function and refactor usages

flferretti:
This PR introduces a `safe_norm` function to handle norm calculations safely, preventing division by zero and ensuring gradient compatibility. 

FYI @CarlottaSartore

----
📚 Documentation preview 📚: https://jaxsim--319.org.readthedocs.build//319/

### DIFF
--- a/src/jaxsim/math/utils.py
+++ b/src/jaxsim/math/utils.py
@@ -6,17 +6,26 @@ import jaxsim.typing as jtp
 def safe_norm(array: jtp.ArrayLike, axis=None) -> jtp.Array:
     """
     Provides a calculation for an array norm so that it is safe
-    to compute the gradient and the NaNs are handled.
+    to compute the gradient and handle NaNs.
 
     Args:
         array: The array for which to compute the norm.
         axis: The axis for which to compute the norm.
+
+    Returns:
+        The norm of the array with handling for zero arrays to avoid NaNs.
     """
 
+    # Check if the entire array is composed of zeros.
     is_zero = jnp.allclose(array, 0.0)
 
+    # Replace zeros with an array of ones temporarily to avoid division by zero.
+    # This ensures the computation of norm does not produce NaNs or Infs.
     array = jnp.where(is_zero, jnp.ones_like(array), array)
 
+    # Compute the norm of the array along the specified axis.
     norm = jnp.linalg.norm(array, axis=axis)
 
+    # Use `jnp.where` to set the norm to 0.0 where the input array was all zeros.
+    # This usage supports potential batch processing for future scalability.
     return jnp.where(is_zero, 0.0, norm)
--- a/src/jaxsim/math/utils.py
+++ b/src/jaxsim/math/utils.py
@@ -1,0 +1,22 @@
+import jax.numpy as jnp
+
+import jaxsim.typing as jtp
+
+
+def safe_norm(array: jtp.ArrayLike, axis=None) -> jtp.Array:
+    """
+    Provides a calculation for an array norm so that it is safe
+    to compute the gradient and the NaNs are handled.
+
+    Args:
+        array: The array for which to compute the norm.
+        axis: The axis for which to compute the norm.
+    """
+
+    is_zero = jnp.allclose(array, 0.0)
+
+    array = jnp.where(is_zero, jnp.ones_like(array), array)
+
+    norm = jnp.linalg.norm(array, axis=axis)
+
+    return jnp.where(is_zero, 0.0, norm)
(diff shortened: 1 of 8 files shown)


src/jaxsim/math/utils.py
```diff
@@ -0,0 +1,22 @@
...
+    to compute the gradient and the NaNs are handled.
+
+    Args:
+        array: The array for which to compute the norm.
+        axis: The axis for which to compute the norm.
+    """
+
+    is_zero = jnp.allclose(array, 0.0)
+
+    array = jnp.where(is_zero, jnp.ones_like(array), array)
```
xela-95, line 18: I'm not entirely sure that is sound to replace a zero array with one full of ones 🤔

flferretti: This is a weird way to allow JAX to still be able to compute the norm, even if this will be discarded as if `is_zero` is True, then in:

```python
    return jnp.where(is_zero, 0.0, norm)
```

The return value would be always 0.0, which is the correct norm of an array of zeros. 
Did that answer you question?

xela-95: Ok understood thanks! using a condition to directly return `0.0` if `is_zero==true` will slow down the code due to branching?

flferretti: You would probably get a TracerBoolConversionError because you are using a plain `if` statement

xela-95: even with a ['jax.lax.cond'](https://jax.readthedocs.io/en/latest/_autosummary/jax.lax.cond.html)?

flferretti: You can use a `jax.lax.cond`, but `jnp.where` already uses `jax.lax.select` under the hood (https://github.com/jax-ml/jax/blob/5031b6f599421bb1c10a1ee045067dc897c2f3a6/jax/_src/numpy/util.py#L272). 

The advantage of using `jnp.where` is that the method would be compatible with batched quantities:

```python
import jax.numpy as jnp
import jaxsim.math

batch_size = 128
quaternions = jnp.array([jnp.ones(4)] * batch_size).at[0].set(jnp.ones(4))

# Check the batched quaternions shape.

print(quaternions.shape)
# >>> (128, 4)

# Check the batched norms shape.

print(jaxsim.math.safe_norm(quaternions, axis=-1).shape)
# >>> (128,)

# Check the batched norm output.

print(jaxsim.math.safe_norm(quaternions, axis=-1))
#  >>> [0. 2. 2. 2. 2. 2. 2. 2. 2. 2. 2. 2. 2. 2. 2. 2. 2. 2. 2. 2. 2. 2. 2. 2. 2. 2. 2. 2. 2. 2. 2. 2. 2. 2. 2. 2. 2. 2. 2. 2. 2. 2. 2. 2. 2. 2. 2. 2. 2. 2. 2. 2. 2. 2. 2. 2. 2. 2. 2. 2. 2. 2. 2. 2. 2. 2. 2. 2. 2. 2. 2. 2. 2. 2. 2. 2. 2. 2. 2. 2. 2. 2. 2. 2. 2. 2. 2. 2. 2. 2. 2. 2. 2. 2. 2. 2. 2. 2. 2. 2. 2. 2. 2. 2. 2. 2. 2. 2. 2. 2. 2. 2. 2. 2. 2. 2. 2. 2. 2. 2. 2. 2. 2. 2. 2. 2. 2. 2.]
```

This is because with `jax.lax.select`, called by `jnp.where` both branches can be executed. I haven't really tested this but I believe that using `jax.lax.cond` would produce an XLA intermediate representation (IR) that expects to execute always the same branch, which is not the case if one of the quaternions to be normalized is full of zeros, leading to errors or not optimized code.

From [JAX documentation](https://jax.readthedocs.io/en/latest/_autosummary/jax.lax.cond.html#jax.lax.cond):

> In contrast with [jax.lax.select()](https://jax.readthedocs.io/en/latest/_autosummary/jax.lax.select.html#jax.lax.select), using cond indicates that only one of the two branches is executed (up to compiler rewrites and optimizations). However, when transformed with [vmap()](https://jax.readthedocs.io/en/latest/_autosummary/jax.vmap.html#jax.vmap) to operate over a batch of predicates, cond is converted to [select()](https://jax.readthedocs.io/en/latest/_autosummary/jax.lax.select.html#jax.lax.select).

xela-95: Ok understood! Then it's ok for me!

xela-95 (approved):
Thanks @flferretti ! For me it's good to go! If you have time, maybe adding a small unit test on the `safe_norm` function for future regression could be a good idea :)

src/jaxsim/math/utils.py
```diff
@@ -0,0 +1,22 @@
...
+
+    norm = jnp.linalg.norm(array, axis=axis)
```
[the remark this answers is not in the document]
flferretti: Thanks! Done in acbe8b7

PR #325, opened 2025-01-04
Slightly optimize computation for `RigidContacts` model

flferretti:
This PR slightly enhances the performances of the `RigidContacts` model by using the pseudoinverse instead of the least-squares method to compute the impact velocity, and removes some unnecessary context managers and assertions

### DIFF
--- a/src/jaxsim/api/model.py
+++ b/src/jaxsim/api/model.py
@@ -2292,9 +2292,6 @@ def step(
         # Hence, here we need to reset the velocity after each impact to guarantee that
         # the linear velocity of the active collidable points is zero.
         case jaxsim.rbda.contacts.RigidContacts():
-            assert isinstance(
-                data_tf.contacts_params, jaxsim.rbda.contacts.RigidContactsParams
-            )
 
             # Raise runtime error for not supported case in which Rigid contacts and
             # Baumgarte stabilization are enabled and used with ForwardEuler integrator.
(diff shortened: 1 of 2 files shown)


CarlottaSartore:
We checked with @flferretti  and the time needed to perform `jnp.linalg.pinv` and `jnp.linalg.lstsq`  are comparable but since `lstq` is more stable for ill conditioned matrix , it could have sense to keep the original implementation. 

Maybe we can close this PR then @flferretti  ?

xela-95 (changes requested):
Thanks @flferretti ! I left a comment

xela-95:
> We checked with @flferretti and the time needed to perform `jnp.linalg.pinv` and `jnp.linalg.lstsq` are comparable but since `lstq` is more stable for ill conditioned matrix , it could have sense to keep the original implementation.
> 
> Maybe we can close this PR then @flferretti ?

Sorry @CarlottaSartore I skipped your comment! Mmmh yes, then maybe we can close this for now. Do you agree @flferretti ?

flferretti:
> Sorry @CarlottaSartore I skipped your comment! Mmmh yes, then maybe we can close this for now. Do you agree @flferretti ?



> Maybe we can close this PR then @flferretti ?

Sure! I also removed a switch of the velocity representation and refactored a bit the function. If you agree I can restore the `jnp.linalg.solve` and keep the rest as it is

xela-95 (approved):

PR #340, opened 2025-01-15
Cache kinematics and dynamics computations in  `JaxSimModelData` 

flferretti:
This PR aims to improve the performances by caching computations and simplifying the codebase. 

The class `JaxSimModelData` will have two additional attributes:

- `_kin_dyn`: private attribute to store the kyndyn computations
- `_kin_dyn_stale`: the cache invalidation flag

The class `KynDynComputation` inherits from `common.ModelDataWithVelocityRepresentation`, so it will have its own `velocity_representation` attribute, which will be useful for performing the conversions.

The kinematics and dynamics computations will be accessible via a property of `JaxSimModelData`, through `KynDynProxy` which is a proxy class that ensures that the attributes of `KynDynComputation` are set and returned in the correct velocity representation.

The value of the cache invalidation flag `JaxSimModelData._kyn_dyn_stale` is set to `True` when any `reset_*` method is called. For this reason, a `RuntimeError` will be raised if the user tries to access to the invalidated cache. At this scope, the `update_kyn_dyn` method would set the cache invalidation flag to `False` and update the `_kyn_dyn` attribute.

----
📚 Documentation preview 📚: https://jaxsim--340.org.readthedocs.build//340/


flferretti:
Closing in favor of #348

PR #342, opened 2025-01-15
Add missing example in documentation thumbnails and update RST file structure

flferretti:
This PR adds a missing example and improves the structure of thumbnail RST file for clarity


xela-95 (approved):
Thanks @flferretti !

PR #343, opened 2025-01-15
Avoid conflicting `pydocstyle` rule in `ruff` configuration

flferretti:
D203 and D211 are incompatible (see https://docs.astral.sh/ruff/rules/incorrect-blank-line-before-class/). Choosing D211


xela-95 (approved):
Agree on the D211! LGTM

PR #344, opened 2025-01-15
Use `jax.block_until_ready` in benchmarks

xela-95:
In this PR I added two little changes:
- Use different random data for each batch (created from different PRNG keys) 
- Use [`jax.block_until_ready`](https://jax.readthedocs.io/en/latest/_autosummary/jax.block_until_ready.html) to measure the time of the full computation and not only the asynchronous dispatch (see https://jax.readthedocs.io/en/latest/async_dispatch.html#async-dispatch)

----
📚 Documentation preview 📚: https://jaxsim--344.org.readthedocs.build//344/

### DIFF
--- a/tests/test_benchmark.py
+++ b/tests/test_benchmark.py
@@ -27,7 +27,10 @@ def benchmark_test_function(
 
     # Warm-up call to avoid including compilation time
     jax.vmap(func, in_axes=(None, 0))(model, data)
-    benchmark(jax.vmap(func, in_axes=(None, 0)), model, data)
+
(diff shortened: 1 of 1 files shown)


xela-95:
> Thanks a lot @xela-95, this was very much needed! LGTM
> 
> Before merging, would you mind updating the PR title to be more specific to the changes, like "Use `jax.block_until_ready` in benchmarks" or something related to avoid measuring the dispatch? Thank you

Done! Rebasing and merging then :)

PR #355, opened 2025-01-28
Avoid device-to-host data transfers and disable runtime exception handling

flferretti:
This PR removes device-to-host data transfers caused by accessing `JaxSimModel` properties and disables runtime exceptions by default. Moreover, it updates the relevant documentation section 

----
📚 Documentation preview 📚: https://jaxsim--355.org.readthedocs.build//355/


younik:
It is indeed

PR #358, opened 2025-01-28
Enable simultaneous visualization of multiple robots in the passive viewer recorder

flferretti:
This PR introduces functionality for simultaneously visualizing multiple robots in the MuJoCo passive viewer recorder

----
📚 Documentation preview 📚: https://jaxsim--358.org.readthedocs.build//358/

### DIFF
--- a/src/jaxsim/mujoco/visualizer.py
+++ b/src/jaxsim/mujoco/visualizer.py
@@ -46,7 +46,7 @@ class MujocoVideoRecorder:
 
         self.fps = fps
         self.frames: list[npt.NDArray] = []
-        self.data: mujoco.MjData | None = None
+        self.data: list[mujoco.MjData] | mujoco.MjData | None = None
         self.model: mujoco.MjModel | None = None
         self.reset(model=model, data=data)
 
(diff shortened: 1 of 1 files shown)


src/jaxsim/mujoco/visualizer.py
```diff
@@ -46,7 +46,7 @@ def __init__(
 
         self.fps = fps
         self.frames: list[npt.NDArray] = []
-        self.data: mujoco.MjData | None = None
+        self.data: list[mujoco.MjData] | mujoco.MjData | None = None
```
younik, line 49: Why is there `data` as init argument, which is not used?
Should we change the type also there?

xela-95 (approved):
Super cool! Thanks @flferretti

src/jaxsim/mujoco/visualizer.py
```diff
@@ -68,8 +68,23 @@ def reset(
     def render_frame(self, camera_name: str = "track") -> npt.NDArray:
         """Render a frame."""
 
-        mujoco.mj_forward(self.model, self.data)
-        self.renderer.update_scene(data=self.data, camera=camera_name)
+        for idx, data in enumerate(
+            self.data if isinstance(self.data, list) else [self.data]
```
younik, line 72: Very minor, but I find more clean to handle it during init/reset, and assuming elsewhere that self.data is always a list.

src/jaxsim/mujoco/visualizer.py
```diff
@@ -68,8 +68,23 @@ def reset(
     def render_frame(self, camera_name: str = "track") -> npt.NDArray:
```
younik, line 68: I cannot comment above lines as you didn't change them, but we should also update the signature of reset method

younik (changes requested):
Thanks @flferretti!

I just left a couple of comments.

Also, in the notebook you should replace `reset_base_position` with a replace call (or set it during build)

src/jaxsim/mujoco/visualizer.py
flferretti: You're right, thanks! Addressed in https://github.com/ami-iit/jaxsim/pull/358/commits/4bc090943033e9006db2ee2ea490351155da1802

src/jaxsim/mujoco/visualizer.py
flferretti: Done in https://github.com/ami-iit/jaxsim/pull/358/commits/4bc090943033e9006db2ee2ea490351155da1802, thanks!

src/jaxsim/mujoco/visualizer.py
younik: ups, yes got it, thanks!

younik (approved):

PR #359, opened 2025-01-30
[Sprint] Improve caching mechanism

younik:
This PR goes a bit beyond what we agreed, so feel free to reject it. I can easily do a simpler version where we call `update_cached` inside the reset function.

The motivation is that update cached was always called after `data.replace` (and the cached values are needed at every step), so it probably makes sense to compute it during data construction and at every replace.

The `reset_` methods are calling a replace, which is an expensive operation. This is not immediately clear to the user, and leads to sequentially reset calls:
```
data.reset_base_position(...)
data.reset_base_velocity(...)
...
```

One method that allows to change multiple things would be better, so I used the replace itself.

The other selling point is that with the current solution, we need to remember to call `update_cached` from now on, which may lead to future bugs.

One negative aspect of this is that we need `model` in `replace`, which differs from the `dataclass.replace` as the model is not an attribute of data.

----
📚 Documentation preview 📚: https://jaxsim--359.org.readthedocs.build//359/


src/jaxsim/api/data.py
```diff
@@ -549,6 +446,64 @@ def reset_base_velocity(
             base_linear_velocity=W_v_WB[0:3].squeeze().astype(float),
             base_angular_velocity=W_v_WB[3:6].squeeze().astype(float),
         )
+    
+    def replace(
```
flferretti, line 450: Adding `typing.override` decorator would prevent in the future errors related to overriding methods from parent class
```suggestion
    @override
    def replace(
```

flferretti (changes requested):
Thanks a lot @younik, it looks to me overall. It does make sense to remove the `reset_*` methods, but the problem is that we are storing some quantities in `VelRepr.Inertial`, so we either need to add a comment to make this clear to the user, or to change the cached quantities in the same velocity representation of the `JaxSimModelData` instance

src/jaxsim/api/link.py
```diff
@@ -295,7 +295,8 @@ def jacobian(
             B_J_WL_I = B_J_WL_B
 
         case VelRepr.Mixed:
-            W_R_B = data.base_orientation(dcm=True)
+            W_R_B = data.base_orientation
```
xela-95, line 298: now data.base_orientation return a quaternion

```suggestion
            W_Q_B = data.base_orientation
```

src/jaxsim/api/data.py
```diff
@@ -550,6 +477,73 @@ def reset_base_velocity(
...
+            joint_positions=joint_positions, base_transform=base_transform
+        )
+        link_transforms, link_velocities = jaxsim.rbda.forward_kinematics_model(
+            model=model,
+            base_position=base_position,
+            base_quaternion=base_quaternion,
+            joint_positions=joint_positions,
+            joint_velocities=joint_velocities,
+            base_linear_velocity=base_linear_velocity,
+            base_angular_velocity=base_angular_velocity,
```
xela-95, line 530: forward kinematics works only in inertial (this has been made more clear in #361 , so we need to convert base linear and angular velocities to this representation before calling it.

xela-95 (changes requested):
Thanks @younik ! The replace approach in which to update the cached values is elegant to me and avoid potential future issues! I added some comments related to the handling of data attributes and their representation.

CarlottaSartore (approved):

src/jaxsim/api/data.py
younik: done!

src/jaxsim/api/link.py
younik: done!

flferretti (approved):

src/jaxsim/api/data.py
younik: good catch, done!

xela-95:
@younik I noted that here we also miss in this PR also a property for the base transform, since it is only cached as a private attribute.

xela-95 (approved):

PR #360, opened 2025-01-30
Reintroduce alternative contact models and streamline changes to the new API

flferretti:
This PR restores the alternative contact models of the original API and streamlines changes to the rest of the codebase

----
📚 Documentation preview 📚: https://jaxsim--360.org.readthedocs.build//360/


flferretti:
Ready for review @ami-iit/darwin. Thanks!

src/jaxsim/api/model.py
```diff
@@ -2079,8 +2104,47 @@ def step(
...
+                    generalized_velocity=BW_ν_pre_impact,
+                    inactive_collidable_points=(δ <= 0),
+                    M=M,
+                    J_WC=J_WC,
+                )
+            )
+
+            # Reset the generalized velocity.
+            data_tf = data_tf.reset_base_velocity(BW_ν_post_impact[0:6])
+            data_tf = data_tf.reset_joint_velocities(BW_ν_post_impact[6:])
```
xela-95, line 2146: Maybe in a future PR we can factor this out in a separate function to avoid cluttering the step function

src/jaxsim/api/model.py
```diff
@@ -246,8 +249,8 @@ def build(
             time_step=time_step,
             terrain=terrain,
             contact_model=contact_model,
-            contacts_params=contacts_params,
-            gravity=gravity,
+            contact_params=contact_params,
+            gravity=-gravity,
```
xela-95, line 253: Maybe it could be more clear to put the minus sign directly where we use the gravity constant inside the function? It's a bit strange to pass `-gravity` to the function invocation

xela-95 (changes requested):
Thanks @flferretti !! I left some minor comments. By the way, remember to remove the `hunt_crossley_model` function from the relaxed rigid contact model since now we have restored the soft contact model that has this function built in.

xela-95:
@flferretti did you notice that here we have also the `test_box_with_zero_gravity` failing?

https://github.com/ami-iit/jaxsim/actions/runs/13061287183/job/36444671783?pr=360#step:11:684

flferretti:
> @flferretti did you notice that here we have also the `test_box_with_zero_gravity` failing?
> 
> 
> 
> https://github.com/ami-iit/jaxsim/actions/runs/13061287183/job/36444671783?pr=360#step:11:684

Yes, but the results are not deterministic, as the test passes if re-run with `pytest --lf`

src/jaxsim/api/contact.py
```diff
@@ -37,14 +37,11 @@ def collidable_point_kinematics(
         the linear component of the mixed 6D frame velocity.
     """
 
-    # Switch to inertial-fixed since the RBDAs expect velocities in this representation.
-    with data.switch_velocity_representation(VelRepr.Inertial):
```
xela-95, line 41: We should pay attention to this change of representation for PR #359 in which the property returns data in the active representation (not always inertial)

src/jaxsim/api/model.py
flferretti: I totally agree! Leaving this unresolved for now

src/jaxsim/api/model.py
flferretti: The fact is that the user might pass a different gravity value, so what we could do is:

```python
gravity = -gravity

# [...]

return cls(..., gravity=gravity)
```

Which is slightly less ugly

src/jaxsim/api/contact.py
younik: Yes, I agree, it should be equivalent

younik (changes requested):
Thank you! I added some comments :)

src/jaxsim/api/model.py
```diff
@@ -2086,6 +2111,54 @@ def step(
...
+                model=model,
+                base_linear_velocity=BW_ν_post_impact[0:3],
+                base_angular_velocity=BW_ν_post_impact[3:6],
+                joint_velocities=BW_ν_post_impact[6:],
+            )
+
+    # Restore the input velocity representation
+    data_tf = data_tf.replace(
+        velocity_representation=data.velocity_representation, validate=False
     )
```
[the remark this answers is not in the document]
flferretti: You're right! Solved in https://github.com/ami-iit/jaxsim/commit/2f1a59e492eff39a777c8deba15fdddc09c77b93

src/jaxsim/api/contact.py
xela-95: yep true. Coming from other languages I don't like to access directly the "private" attributes of a class, is like coupling to the inner implementation of it. If in a next PR we change the representation in which they are stored it will be difficult to backtrace errors here, while if we use the class API we are relying on the class contract.

younik: Good point, we should see them as _protected_ rather than _private_

src/jaxsim/api/contact.py
```diff
@@ -190,8 +196,66 @@ def estimate_good_contact_parameters(
...
+            parameters = contacts.RigidContactsParams.build(
+                mu=static_friction_coefficient,
+                **(
+                    dict(
+                        K=K,
+                        D=2 * jnp.sqrt(K),
+                    )
+                    | kwargs
+                ),
+            )
```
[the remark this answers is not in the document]
flferretti: Done in https://github.com/ami-iit/jaxsim/pull/360/commits/4f0a3f4cf18474c63bc81079ca5c057a2ccefe43, thanks!

src/jaxsim/api/model.py
```diff
@@ -2065,6 +2070,26 @@ def step(
...
+            contact_state["tangential_deformation"] = aux_dict["m_dot"]
+
+        case (
+            jaxsim.rbda.contacts.RigidContacts()
+            | jaxsim.rbda.contacts.RelaxedRigidContacts()
+        ):
+            pass
+
+        case _:
+            raise ValueError(f"Invalid contact model: {model.contact_model}")
```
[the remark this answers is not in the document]
flferretti: Done in https://github.com/ami-iit/jaxsim/pull/360/commits/07d6eca03a4ba2f6852c3c2806fbb88a1ce381b5, thanks!

src/jaxsim/api/model.py
```diff
@@ -447,6 +450,8 @@ def reduce(
         time_step=model.time_step,
         terrain=model.terrain,
         contact_model=model.contact_model,
+        contact_params=model.contact_params,
+        gravity=-model.gravity,
```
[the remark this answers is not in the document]
flferretti: This is a bit tricky. `model.gravity` is saved as negative, but the `build` method expects to receive a positive constant

src/jaxsim/api/model.py
```diff
@@ -2052,7 +2057,7 @@ def step(
 
         # Compute the 6D forces W_f ∈ ℝ^{n_L × 6} applied to links due to contact
         # with the terrain.
-        W_f_L_terrain = js.contact_model.link_contact_forces(
+        W_f_L_terrain, aux_dict = js.contact_model.link_contact_forces(
```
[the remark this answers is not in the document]
flferretti: As discussed f2f, we could move the `contact_model` module into `rbda.contacts.common`. WDYT @xela-95 @younik?

src/jaxsim/api/data.py
```diff
@@ -183,6 +198,7 @@ def build(
             _joint_transforms=joint_transforms,
             _link_transforms=link_transforms,
             _link_velocities=link_velocities_inertial,
+            contact_state=contact_state or {},
```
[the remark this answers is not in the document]
flferretti: Done in https://github.com/ami-iit/jaxsim/pull/360/commits/2c3153ecb347176d1689636de99cb33e82d59072, thanks!

src/jaxsim/api/data.py
```diff
@@ -171,6 +176,16 @@ def build(
...
 
+        contact_state = (
+            {
+                "tangential_deformation": jnp.zeros_like(
+                    model.kin_dyn_parameters.contact_parameters.point
+                )
+            }
+            if isinstance(model.contact_model, jaxsim.rbda.contacts.SoftContacts)
+            else contact_state or {}
+        )
```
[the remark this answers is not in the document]
flferretti: Addressed in https://github.com/ami-iit/jaxsim/pull/360/commits/2c3153ecb347176d1689636de99cb33e82d59072. Thanks! 

It might be useful when a vectorized data object is created starting from a single data instance

src/jaxsim/api/model.py
flferretti: > May be helpful to also update the doc and let know the user how we expect gravity to be

Done in https://github.com/ami-iit/jaxsim/pull/360/commits/69720754a302fd8f7998af8ca5c688557d69fd1e

src/jaxsim/api/model.py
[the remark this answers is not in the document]
younik: I agree with it

src/jaxsim/api/contact.py
flferretti: Leaving unresolved for now, thanks for pointing this out

younik (approved):

src/jaxsim/api/model.py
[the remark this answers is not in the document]
flferretti: Done in https://github.com/ami-iit/jaxsim/pull/360/commits/7ec23b5e7d05dbfa8e8b433896e2c7964c23c821

CarlottaSartore:
Despite the ongoing PR review process, Before merging this, I would run a no-regression test for the time performances and also `comodo` walking

xela-95 (approved):

src/jaxsim/api/model.py
```diff
@@ -2073,13 +2076,16 @@ def step(
 
         # Compute the 6D forces W_f ∈ ℝ^{n_L × 6} applied to links due to contact
         # with the terrain.
-        W_f_L_terrain = js.contact_model.link_contact_forces(
+        W_f_L_terrain, old_contact_state = js.contact.link_contact_forces(
```
CarlottaSartore, line 2079: I would add a comment here explaining why we are calling a variable "old contact state" since could be tricky to recall in the near future

CarlottaSartore (changes requested):
Before merging we need to: 
- Check if the robot still walks in comodo with all the contacts (as it was before the sprint)
- Check  if the compilation and run time  to be sure we did not introduce any slow-down factor in this PR 

For the rest it is a go !

flferretti:
Thanks for the review! 

> * Check if the robot still walks in comodo with all the contacts (as it was before the sprint)

Done. I'll report here the different performances achieved:

| Contact Model | RTF|
|--------|--------|
| Soft | 35.9% |
| RelaxedRigid | 20.9% | 
| Rigid | WIP |

The rigid contact model seems to have a long compilation time. To be discussed with @xela-95 tomorrow

> * Check if the compilation and run time to be sure we did not introduce any slow-down factor in this PR

Each contact model has its own logic and its own computation time. The PR doesn’t change how the default model runs, so a slow-down wouldn't be possible unless someone explicitly switches to a different model, which is expected to have a different speed.

Let me know if there's a specific concern beyond that

CarlottaSartore:
> Each contact model has its own logic and its own computation time. The PR doesn’t change how the default model runs, so a slow-down wouldn't be possible unless someone explicitly switches to a different model, which is expected to have a different speed.
> 
> Let me know if there's a specific concern beyond that

We are not timing from a lot of PRs that have been merged waiting for an analysis on performances (see https://github.com/ami-iit/jaxsim/pull/378#issuecomment-2684358253) so this is something that needs to be done. The computation did not change even during the sprint but we speed up things around by changing how things were called. Once https://github.com/ami-iit/jaxsim/pull/380 is finished we will not have to worry on these things.  

Before merging you can use either this PR when finished https://github.com/ami-iit/jaxsim/pull/380 to evaluate the time performances or if you prefer you can do that on your machine.

  Indeed 
  <html>
<body>

Contact Model | RTF
-- | --
Soft | 35.9%
RelaxedRigid | 20.9%
Rigid | WIP


</body>
</html>
  Cool that we did not introduce any regression for comodo! For the time information, it is difficult to interpret without a reference to the main or information on the machine.

flferretti:
> > Each contact model has its own logic and its own computation time. The PR doesn’t change how the default model runs, so a slow-down wouldn't be possible unless someone explicitly switches to a different model, which is expected to have a different speed.
> 
> > 
> 
> > Let me know if there's a specific concern beyond that
> 
> 
> 
> We are not timing from a lot of PRs that have been merged waiting for an analysis on performances (see https://github.com/ami-iit/jaxsim/pull/378#issuecomment-2684358253) so this is something that needs to be done. The computation did not change even during the sprint but we speed up things around by changing how things were called. Once https://github.com/ami-iit/jaxsim/pull/380 is finished we will not have to worry on these things.  
> 
> 
> 
> Before merging you can use either this PR when finished https://github.com/ami-iit/jaxsim/pull/380 to evaluate the time performances or if you prefer you can do that on your machine.
> 
> 
> 
> | Contact Model | RTF|
> 
> |--------|--------|
> 
> | Soft | 35.9% |
> 
> | RelaxedRigid | 20.9% | 
> 
> | Rigid | WIP |
> 
> 
> 
> 
> 
>   Cool that we did not introduce any regression for comodo! For the time information, it is difficult to interpret without a reference to the main or information on the machine.
> 
> 

We actually did check in #378. The other merged PRs did not affect any computations, for this reason the performances were not tested. Furthermore, we are intensively using jaxsim everyday, so any significant change would be very evident. 

I wouldn't use comodo for quantitative benchmarking as it involves many different libraries and CPU-only computations. As long as we are staying in the same order of magnitude, I can consider myself to be happy with the results. 

The computation did change in the sprint. We were heavily relying on  a non-functional approach and intermediate computations were not cached. It is definitely not the case of this PR.

CarlottaSartore:
As discussed F2f with @flferretti  we will do before #380  and then use that for evaluating main and this PR before merging

flferretti:
> > # Benchmark
> > This comment was automatically generated by [workflow](https://github.com/ami-iit/jaxsim/actions?query=workflow%3AGPU%20Benchmarks) using [github-action-benchmark](https://github.com/marketplace/actions/continuous-benchmark).
> 
> Notice that the benchmark shows a 5x degradation in perf on some test. (Will this be raised by the bot itself after last commit?)

Yes, this may depend on various factor, among which the current GPU utilization of the runner. I'll further investigate this in the next days

flferretti:
@younik @xela-95 @CarlottaSartore, `test_benchmark.py::test_relaxed_rigid_contact_model` is taking more time as before `system_acceleration` did not include the computation of contact forces and now it does. It was something that was removed during the sprint and after restoring the benchmarks I did not notice that there wasn't actually any call to `compute_contact_forces`.

Ready for review then, thank you

src/jaxsim/rbda/contacts/common.py
```diff
@@ -173,3 +255,99 @@ def _parameters_class(cls) -> type[ContactsParams]:
...
+        # Import the contact models to avoid circular imports.
+        from .relaxed_rigid import RelaxedRigidContacts
+        from .rigid import RigidContacts
+        from .soft import SoftContacts
+
+        match self:
+            case SoftContacts():
+                return {"tangential_deformation": old_contact_state["m_dot"]}
+            case RigidContacts() | RelaxedRigidContacts():
+                return {}
```
younik, line 281: Can we use polymorphism for this?

src/jaxsim/rbda/contacts/common.py
```diff
@@ -173,3 +255,99 @@ def _parameters_class(cls) -> type[ContactsParams]:
...
+                in_axes=(0, 0, None),
+            )(W_p_C, jnp.zeros_like(W_p_C), model.terrain)
+
+            with data.switch_velocity_representation(VelRepr.Mixed):
+                J_WC = js.contact.jacobian(model, data)[
+                    indices_of_enabled_collidable_points
+                ]
+                M = js.model.free_floating_mass_matrix(model, data)
+                BW_ν_pre_impact = data.generalized_velocity
+
```
younik, line 326: Same here, can we use polymorphism instead?

src/jaxsim/rbda/contacts/soft.py
```diff
@@ -0,0 +1,417 @@
...
+                HashedNumpyArray.hash_of_array(self.mu),
+                HashedNumpyArray.hash_of_array(self.p),
+                HashedNumpyArray.hash_of_array(self.q),
+            )
+        )
+
+    def __eq__(self, other: SoftContactsParams) -> bool:
+
+        if not isinstance(other, SoftContactsParams):
+            return NotImplemented
```
younik, line 65: it should return False here(?)

younik (commented):
Good job! Just a couple of more comments :)

src/jaxsim/rbda/contacts/common.py
flferretti: Can you be more specific?

younik: I mean to have this function being a no-op in the baseclass, and override it in the RigidContacts class

src/jaxsim/rbda/contacts/common.py
younik: while here having it as abstract class implement the logics in the subclasses

flferretti: Got it, thanks! Done in https://github.com/ami-iit/jaxsim/pull/360/commits/62f02b459abfad7b379916ecc9f6364f12d6a2a3

src/jaxsim/rbda/contacts/common.py
flferretti: Done in https://github.com/ami-iit/jaxsim/pull/360/commits/62f02b459abfad7b379916ecc9f6364f12d6a2a3

younik (approved):

CarlottaSartore:
Thanks @flferretti, I was looking at this comment with @xela-95

> @younik @xela-95 @CarlottaSartore, test_benchmark.py::test_relaxed_rigid_contact_model is taking more time as before system_acceleration did not include the computation of contact forces and now it does. It was something that was removed during the sprint and after restoring the benchmarks I did not notice that there wasn't actually any call to compute_contact_forces. For this reason, the failure on the benchmark is expected

And we had some doubts, can you  please explain better what did happen ? especially for 

>  I did not notice that there wasn't actually any call to compute_contact_forces. 

Was it a bug in the actual code implementation of something specifically related to the benchmark ?

flferretti:
> And we had some doubts, can you please explain better what did happen ? especially for

Yes, sure! So basically, originally in the benchmark tests we were benchmarking the call to `js.ode.system_dynamics`, which includes a call to `js.ode.system_acceleration` and `js.ode.system_position_dynamics`. The issue was that during the sprint, in https://github.com/ami-iit/jaxsim/commit/2d3b3ae945cf076c4b9c4aedb611cef191c0d681, we removed the call the computation of the contact forces. For this reason, the benchmark was just computing the acceleration without actually calling the contact model. That's why the computational time required now is larger. Let me know if anything is still unclear

PR #373, opened 2025-02-14
Implement Runge-Kutta 4 integrator

flferretti:
This PR introduces a new Runge-Kutta 4 integration method and an `Integrator` enum to select between available simulation integrators. Moreover, it adds NaN checks for quaternions and add a `pytest` fixture for testing different integrators and updates the `JaxSimModel` to support integrator selection.

----
📚 Documentation preview 📚: https://jaxsim--373.org.readthedocs.build//373/

### DIFF
--- a/src/jaxsim/api/integrators.py
+++ b/src/jaxsim/api/integrators.py
@@ -1,4 +1,5 @@
 import dataclasses
+from collections.abc import Callable
 
 import jax
 import jax.numpy as jnp
@@ -143,3 +144,9 @@ def rk4_integration(
     data_tf = dataclasses.replace(data, **{"_" + k: v for k, v in x_tf.items()})
 
     return data_tf.replace(model=model)
+
+
+_INTEGRATORS_MAP: dict[js.model.Integrator, Callable[..., js.data.JaxSimModelData]] = {
+    js.model.Integrator.SemiImplicitEuler: semi_implicit_euler_integration,
+    js.model.Integrator.RungeKutta4: rk4_integration,
+}
--- a/src/jaxsim/api/integrators.py
+++ b/src/jaxsim/api/integrators.py
@@ -90,30 +90,11 @@ def rk4_integration(
 
     dt = model.time_step
 
-    def get_state_derivative(data_ode: JaxSimModelData) -> dict:
-
-        # Safe normalize the quaternion.
-        base_quaternion_norm = jaxsim.math.safe_norm(data_ode._base_quaternion)
-        base_quaternion = data_ode._base_quaternion / jnp.where(
-            base_quaternion_norm == 0, 1.0, base_quaternion_norm
-        )
-
-        return dict(
-            base_position=data_ode._base_position,
-            base_quaternion=base_quaternion,
-            joint_positions=data_ode._joint_positions,
-            base_linear_velocity=data_ode._base_linear_velocity,
-            base_angular_velocity=data_ode._base_angular_velocity,
-            joint_velocities=data_ode._joint_velocities,
-        )
-
     def f(x) -> dict[str, jtp.Matrix]:
 
         with data.switch_velocity_representation(jaxsim.VelRepr.Inertial):
 
-            data_ti = data.replace(
-                model=model, **{k.lstrip("_"): v for k, v in x.items()}
-            )
+            data_ti = data.replace(model=model, **x)
 
             return js.ode.system_dynamics(
                 model=model,
@@ -122,7 +103,19 @@ def rk4_integration(
                 joint_torques=joint_torques,
             )
 
-    x_t0 = get_state_derivative(data)
+    base_quaternion_norm = jaxsim.math.safe_norm(data._base_quaternion)
+    base_quaternion = data._base_quaternion / jnp.where(
+        base_quaternion_norm == 0, 1.0, base_quaternion_norm
+    )
+
+    x_t0 = dict(
+        base_position=data._base_position,
+        base_quaternion=base_quaternion,
+        joint_positions=data._joint_positions,
+        base_linear_velocity=data._base_linear_velocity,
+        base_angular_velocity=data._base_angular_velocity,
+        joint_velocities=data._joint_velocities,
+    )
 
     euler_mid = lambda x, dxdt: x + (0.5 * dt) * dxdt
     euler_fin = lambda x, dxdt: x + dt * dxdt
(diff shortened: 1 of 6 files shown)


src/jaxsim/api/integrators.py
```diff
@@ -74,3 +76,77 @@ def semi_implicit_euler_integration(
...
+            joint_velocities=data_ode.joint_velocities,
+        )
+
+    def f(x) -> dict[str, jtp.Matrix]:
+
+        with data.switch_velocity_representation(jaxsim.VelRepr.Inertial):
+
+            data_ti = data.replace(
+                model=model, **{k.lstrip("_"): v for k, v in x.items()}
+            )
```
younik, line 116: I am afraid this will make data_ti in inertial (it never exits from the with, as we discussed previously).

If you need to put some representation in inertial, you can use dataclasses.replace on the protected attributes (that are in intertial)

src/jaxsim/api/integrators.py
```diff
@@ -74,3 +76,77 @@ def semi_implicit_euler_integration(
...
+
+            return js.ode.system_dynamics(
+                model=model,
+                data=data_ti,
+                link_forces=link_forces,
+                joint_torques=joint_torques,
+            )
+
+    with data.switch_velocity_representation(jaxsim.VelRepr.Inertial):
+        x_t0 = get_state_derivative(data)
```
younik, line 126: This switch is avoidable maybe if you access the protected attributes inside the get_state_derivative function. (not big deal)

src/jaxsim/api/integrators.py
```diff
@@ -74,3 +76,77 @@ def semi_implicit_euler_integration(
...
+
+    # Average the slopes and compute the RK4 state derivative.
+    average = lambda k1, k2, k3, k4: (k1 + 2 * k2 + 2 * k3 + k4) / 6
+
+    dxdt = jax.tree_util.tree_map(average, k1, k2, k3, k4)
+
+    # Integrate the dynamics
+    x_tf = jax.tree_util.tree_map(euler_fin, x_t0, dxdt)
+
+    data_tf = dataclasses.replace(data, **{"_" + k: v for k, v in x_tf.items()})
```
younik, line 144: personal opinion: I would list the attributes, as I find it more readable, even if less concise

src/jaxsim/api/model.py
```diff
@@ -23,6 +24,13 @@
 from .common import VelRepr
 
 
+class Integrator(enum.IntEnum):
```
younik, line 27: Again, just a matter of personal opinion: "IntegratorType" sounds a better name, as Integrator would suggest to me that this is an Integrator class, i.e. implementing integrators (similar as before)

younik (commented):
Good job, thanks!
I added a couple of minor comments

src/jaxsim/api/integrators.py
flferretti: Yes, that is the intended behavior

src/jaxsim/api/integrators.py
flferretti: Right! Thanks

flferretti: Addressed in https://github.com/ami-iit/jaxsim/pull/373/commits/21b5ecbd3c9058faefc198f3c8534c31fe09658a

src/jaxsim/api/model.py
flferretti: Addressed in IntegratorType

src/jaxsim/api/integrators.py
younik: Ah okay, I see it now, we still use data later, thanks

src/jaxsim/api/integrators.py
```diff
@@ -74,3 +76,78 @@ def semi_implicit_euler_integration(
...
+            )
+
+            return js.ode.system_dynamics(
+                model=model,
+                data=data_ti,
+                link_forces=link_forces,
+                joint_torques=joint_torques,
+            )
+
+    x_t0 = get_state_derivative(data)
```
xela-95, line 125: Why is the function called get_state_derivative? Doesn't that just return the input data with the normalized quaternion?

xela-95 (approved):
Thanks! I really like the simplicity of the implementation! Left some minor comments

src/jaxsim/api/ode.py
```diff
@@ -158,14 +158,11 @@ def system_dynamics(
...
-    ode_state_derivative = JaxSimModelData.build(
-        model=model,
+    return dict(
         base_position=W_ṗ_B,
         base_quaternion=W_Q̇_B,
         joint_positions=ṡ,
         base_linear_velocity=W_v̇_WB[0:3],
         base_angular_velocity=W_v̇_WB[3:6],
         joint_velocities=s̈,
     )
```
xela-95, line 168: Is this safe? Returning a dict is not ambiguous here?

flferretti: Do you mean in the sense that the values can be changes? In that case we could use a named tuple or a similar structure. I would avoid using an additional JAX dataclass, like we used to do some months ago. The idea was to make the implementation as simple as possible.
Another option might be to make the function "private" and update the docstrings (my bad, I still need to do it)

src/jaxsim/api/integrators.py
flferretti: You may find a WIP at https://github.com/ami-iit/jaxsim/pull/373/commits/79fec1290dfab08b20b993e59f63644d95064009

src/jaxsim/api/integrators.py
flferretti: Removed in https://github.com/ami-iit/jaxsim/pull/373/commits/79fec1290dfab08b20b993e59f63644d95064009

src/jaxsim/api/ode.py
younik: I like the named tuple option!

younik (approved):

src/jaxsim/api/ode.py
flferretti: Okay, I checked better and neither NamedTuple nor MappingProxyType are valid JAX types. What I could do it to register them as valid PyTree nodes as e.g.:

```python
jax.tree_util.register_pytree_node(MappingProxyType,
    flatten_func=lambda x: list(x.items()),
    unflatten_func=lambda items: MappingProxyType(dict(items))
)
``` 

But I think it's an overkill, what about just leaving it as a dict?

younik: Returning a dict it is okay for me

xela-95: Yep let's proceed :)

src/jaxsim/api/integrators.py
xela-95: Both option 2 and @younik one are good for me :)

flferretti: I'll leave as it is then, thank you both!

PR #374, opened 2025-02-17
Remove object references cycle in LinkDescription

younik:
Currently, `LinkDescription` has the attributes parent and children that are LinkDescription. This creates a reference where a `LinkDescription` references its parent, which references back to the `LinkDescription`.

This can create infinite recursion in case we call a function recursively on the attributes of `LinkDescription recursively`, for example in `__repr__`, `__has__` or `dataclasses.asdict`.

The last one is needed to implement a [Gymnax](https://github.com/RobertTLange/gymnax) environment, which I believe is a common use case.

In this PR, I changed the parent attribute to be the parent's name, instead of the object itself. This doesn't create a problem since most of the time we only need the name (and there is a dictionary mapping names to LinkDescription).

To reproduce the issue:
```python
from dataclasses import asdict
import jaxsim.api as js
import jaxsim

    model_description=...,
    is_urdf=True,
    contact_model=jaxsim.rbda.contacts.RelaxedRigidContacts.build(),
)
asdict(model)
```

### DIFF
--- a/src/jaxsim/parsers/descriptions/link.py
+++ b/src/jaxsim/parsers/descriptions/link.py
@@ -31,7 +31,7 @@ class LinkDescription(JaxsimDataclass):
     mass: float = dataclasses.field(repr=False)
     inertia: jtp.Matrix = dataclasses.field(repr=False)
     index: int | None = None
-    parent: LinkDescription | None = dataclasses.field(default=None, repr=False)
+    parent_name: str | None = dataclasses.field(default=None, repr=False)
     pose: jtp.Matrix = dataclasses.field(default_factory=lambda: jnp.eye(4), repr=False)
 
     children: Static[tuple[LinkDescription]] = dataclasses.field(
@@ -50,8 +50,7 @@ class LinkDescription(JaxsimDataclass):
                 hash(int(self.index)) if self.index is not None else 0,
                 HashedNumpyArray.hash_of_array(self.pose),
                 hash(tuple(self.children)),
-                # Here only using the name to prevent circular recursion:
-                hash(self.parent.name) if self.parent is not None else 0,
+                hash(self.parent_name) if self.parent_name is not None else 0,
             )
         )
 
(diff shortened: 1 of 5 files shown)


src/jaxsim/parsers/kinematic_graph.py
```diff
@@ -971,13 +971,10 @@ def find_parent_link_of_frame(self, name: str) -> str:
...
-            case _:
-                msg = f"Failed to find parent element of frame '{name}' with name '{frame.parent.name}'"
-                raise RuntimeError(msg)
+        if frame.parent_name in self.graph.links_dict:
+            return frame.parent_name
+        elif frame.parent_name in self.graph.frames_dict:
+            return self.find_parent_link_of_frame(name=frame.parent_name)
+        else:
+            msg = f"Failed to find parent element of frame '{name}' with name '{frame.parent_name}'"
+            raise RuntimeError(msg)
```
flferretti, line 980: The `else` can be removed
```suggestion
        if frame.parent_name in self.graph.links_dict:
            return frame.parent_name
        elif frame.parent_name in self.graph.frames_dict:
            return self.find_parent_link_of_frame(name=frame.parent_name)

        msg = f"Failed to find parent element of frame '{name}' with name '{frame.parent_name}'"
        raise RuntimeError(msg)
```

src/jaxsim/parsers/descriptions/link.py
```diff
@@ -31,7 +31,7 @@ class LinkDescription(JaxsimDataclass):
     mass: float = dataclasses.field(repr=False)
     inertia: jtp.Matrix = dataclasses.field(repr=False)
     index: int | None = None
-    parent: LinkDescription | None = dataclasses.field(default=None, repr=False)
+    parent_name: str | None = dataclasses.field(default=None, repr=False)
```
flferretti, line 34: In #351 the `jax_dataclasses` will be removed, but for the time being we should mark the string attribute as static, since it's not a JAX type. This is not a problem for the time being since we don't have any JIT-ted method anyway
```suggestion
    parent_name: Static[str | None] = dataclasses.field(default=None, repr=False)
```

flferretti (changes requested):
Thanks a lot @younik, this is very useful also for #351 and for @xela-95! I just added a couple minor comments

xela-95 (approved):
Thanks! LGTM!

flferretti (approved):

PR #376, opened 2025-02-19
Update `pyproject.toml` to use `depends-on` pixi parameter

xela-95:
----
📚 Documentation preview 📚: https://jaxsim--376.org.readthedocs.build//376/


flferretti (approved):
Thanks @xela-95, LGTM

PR #378, opened 2025-02-25
Uniformize usage of integrators API

flferretti:
This PR changes the API usage for integrators by ensuring consistent parameter names and removing redundant calculations. This improves code clarity and maintainability.

----
📚 Documentation preview 📚: https://jaxsim--378.org.readthedocs.build//378/

### DIFF
--- a/src/jaxsim/api/integrators.py
+++ b/src/jaxsim/api/integrators.py
@@ -8,72 +8,75 @@ import jaxsim
 import jaxsim.api as js
 import jaxsim.typing as jtp
 from jaxsim.api.data import JaxSimModelData
-from jaxsim.math import Adjoint, Transform
+from jaxsim.math import Skew
 
 
 def semi_implicit_euler_integration(
     model: js.model.JaxSimModel,
     data: js.data.JaxSimModelData,
-    base_acceleration_inertial: jtp.Vector,
-    joint_accelerations: jtp.Vector,
+    link_forces: jtp.Vector,
+    joint_torques: jtp.Vector,
 ) -> JaxSimModelData:
     """Integrate the system state using the semi-implicit Euler method."""
-    # Step the dynamics forward.
+
     with data.switch_velocity_representation(jaxsim.VelRepr.Inertial):
 
+        # Compute the system acceleration
+        W_v̇_WB, s̈ = js.ode.system_acceleration(
+            model=model,
+            data=data,
+            link_forces=link_forces,
+            joint_torques=joint_torques,
+        )
+
         dt = model.time_step
-        W_v̇_WB = base_acceleration_inertial
-        s̈ = joint_accelerations
 
-        B_H_W = Transform.inverse(data._base_transform).at[:3, :3].set(jnp.eye(3))
-        BW_X_W = Adjoint.from_transform(B_H_W)
-
+        # Compute the new generalized velocity.
         new_generalized_acceleration = jnp.hstack([W_v̇_WB, s̈])
-
         new_generalized_velocity = (
             data.generalized_velocity + dt * new_generalized_acceleration
         )
 
-        new_base_velocity_inertial = new_generalized_velocity[0:6]
-        new_joint_velocities = new_generalized_velocity[6:]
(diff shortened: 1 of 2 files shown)


younik (approved):
I strongly agree with these changes, thank you!

CarlottaSartore (changes requested):
LGTM but also in this case, berfore mergign I would: 
- Check if the robot walks in comodo with relaxed rigid (it happened  in the past when changing integrator that the tests passed but the robot was not walking)
- Check the time performance

flferretti:
Thanks for your review @CarlottaSartore! Regarding:

>     * Check if the robot walks in comodo with relaxed rigid (it happened  in the past when changing integrator that the tests passed but the robot was not walking)

Only the interface has been changed. There are no significant changes to the integration method apart from using this formulation:

```math
{} ^{B[W]} v _{W, B} =
\begin{bmatrix}
{} ^W \dot{o} _B \\
{} ^W \omega _B
\end{bmatrix} = {} ^W v _{W, B} - \begin{bmatrix} S({} ^W \omega _B) {} ^W o _B \\ 0 \end{bmatrix}
```

Instead of explicitly converting to mixed representation. This approach has already been verified as it is the one used by the RK4 integrator.

>     * Check the time performance

I removed an extra call to the system dynamics for the RK4 as it wasn't used. Since the default contact model has not been changed. Do you think it is really worth at this stage or can we address this in future PRs?

CarlottaSartore:
> Thanks for your review @CarlottaSartore! Regarding:
> 
> > ```
> > * Check if the robot walks in comodo with relaxed rigid (it happened  in the past when changing integrator that the tests passed but the robot was not walking)
> > ```
> 
> Only the interface has been changed. There are no significant changes to the integration method apart from using this formulation:
> 
> 𝐵
> [
> 𝑊
> ]
> 𝑣
> 𝑊
> ,
> 𝐵
> =
> [
> 𝑊
> 𝑜
> ˙
> 𝐵
>  
> 𝑊
> 𝜔
> 𝐵
> ]
> =
> 𝑊
> 𝑣
> 𝑊
> ,
> 𝐵
> −
> [
> 𝑆
> (
> 𝑊
> 𝜔
> 𝐵
> )
> 𝑊
> 𝑜
> 𝐵
> 0
> ]
> Instead of explicitly converting to mixed representation. This approach has already been verified as it is the one used by the RK4 integrator.
> 
> > ```
> > * Check the time performance
> > ```
> 
> I removed an extra call to the system dynamics for the RK4 as it wasn't used. Since the default contact model has not been changed. Do you think it is really worth at this stage or can we address this in future PRs?

I would double-check with Comodo, to be sure that we have replicated it correctly, for what concerns the time analysis I think you can  do one for both this and https://github.com/ami-iit/jaxsim/pull/360,  (merging in a unique PR) the same olds for comodo check  ! Up to you!

flferretti:
The changes correctly run on CoMoDo. Merging

CarlottaSartore:
> The changes correctly run on CoMoDo. Merging

Cool, I assume then that the time performance check will be done in #360,  it would have been better to do it in a unique PR so that we are more careful in merging in main but since we are doing #380 this kind of issues will be automatically correctly handled.

flferretti:
> > The changes correctly run on CoMoDo. Merging
> 
> 
> 
> Cool, I assume then that the time performance check will be done in #360,  it would have been better to do it in a unique PR so that we are more careful in merging in main but since we are doing #380 this kind of issues will be automatically correctly handled.

The performance checks have already been done. The results didn't present any significant difference

CarlottaSartore:
Cool on which machine ? It would be better to report here the performance output as done here https://github.com/ami-iit/component_darwin/issues/59#issuecomment-2639530952 for completeness

flferretti:
> Cool on which machine ? 

Intel Ultra 7 155H. The tests were performed using comodo walking task as a reference. 

> It would be better to report here the performance output as done here https://github.com/ami-iit/component_darwin/issues/59#issuecomment-2639530952 for completeness  

You're right, I'll keep working on #380, so we don't have to do this manually

PR #382, opened 2025-02-27
Exclude overlapping collidable points during collision parsing

flferretti:
This PR implements a logic to filter out overlapping collidable points when extracting their positions, ensuring unique points are processed for contact parameters. This is particularly useful when dealing with closed-loop kinematic chains or ErgoCub models

----
📚 Documentation preview 📚: https://jaxsim--382.org.readthedocs.build//382/

### DIFF
--- a/docs/guide/configuration.rst
+++ b/docs/guide/configuration.rst
@@ -17,6 +17,10 @@ Environment variables starting with ``JAXSIM_COLLISION_`` are used to configure 
 
   *Default:* ``False``.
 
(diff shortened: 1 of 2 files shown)


flferretti:
> LGTM! Do we want it `JAXSIM_COLLISION_INCLUDE_OVERLAPS=False` by default? (Just asking because previously it was working as if it was True)

Thanks @younik! I'm getting some errors in CI which I need to fix, but once fixed, we may consider leaving this on by default as it can improve performances on some models

flferretti:
Closing for the time being as this requires heavy recomputation of the collidable points parent

PR #394, opened 2025-03-27
Enable reverse-mode autodiff for contact solves via gradients override

flferretti:
This pull request includes changes to improve the optimization process in the `relaxed_rigid` module and updates to the `test_automatic_differentiation` tests. The most important changes include replacing the optimization function with a custom linear solver, removing unnecessary imports, and updating the gradient check modes. Additionally, this update enables reverse-mode automatic differentiation by injecting implicit gradients at the contact solve, which also allows the use of non-differentiable GPU-based solvers while maintaining differentiability for parameters optimization.

Improvements to optimization process:

* [`src/jaxsim/rbda/contacts/relaxed_rigid.py`](diffhunk://#diff-4465afc5357a08994f4ce0fce330c32ab819938dbfe2fbb9d4e1974e44d46409L480-R496): Wrapped the `run_optimization` function with `jax.lax.custom_linear_solve` to override gradient computation with implicit differences.

Updates to tests:

* [`tests/test_automatic_differentiation.py`](diffhunk://#diff-d9535bc164ad7c0a31f18f88ad7ccc7c0e615f38b4430b34226a643396024fd4L353-L357): Removed the conditional skip for the `ergoCub` model when `optax` version is less than or equal to 0.2.4.
* [`tests/test_automatic_differentiation.py`](diffhunk://#diff-d9535bc164ad7c0a31f18f88ad7ccc7c0e615f38b4430b34226a643396024fd4L419-R416): Updated the `check_grads` function to include both forward and reverse modes for gradient checking.

----
📚 Documentation preview 📚: https://jaxsim--394.org.readthedocs.build//394/

### DIFF
--- a/src/jaxsim/rbda/contacts/relaxed_rigid.py
+++ b/src/jaxsim/rbda/contacts/relaxed_rigid.py
@@ -477,13 +477,21 @@ class RelaxedRigidContacts(common.ContactModel):
         tol = solver_options.pop("tol")
         maxiter = solver_options.pop("maxiter")
 
-        # Compute the 3D linear force in C[W] frame.
-        solution, _ = run_optimization(
+        solve_fn = lambda *_: run_optimization(
             init_params=init_params,
             fun=objective,
             opt=optax.lbfgs(**solver_options),
             tol=tol,
             maxiter=maxiter,
+        )
+
+        # Compute the 3D linear force in C[W] frame.
+        solution, _ = jax.lax.custom_linear_solve(
+            lambda x: A @ x,
+            -b,
+            solve=solve_fn,
+            symmetric=True,
+            has_aux=True,
         )
 
         # Reshape the optimized solution to be a matrix of 3D contact forces.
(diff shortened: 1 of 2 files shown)


xela-95 (approved):
Looks good to me! 

@flferretti just a curiosity: do the regularization terms introduced in A and b maintain the symmetric structure of the problem? I saw that if this is not the case we should provide also the `transpose_solve` function to `custom_linear_solve`

younik (approved):
Looks good!

flferretti:
> @flferretti just a curiosity: do the regularization terms introduced in A and b maintain the symmetric structure of the problem? I saw that if this is not the case we should provide also the `transpose_solve` function to `custom_linear_solve`

It should, as `R` is a diagonal regularization matrix. `transpose_solve` is necessary only if `symmetric` is not set:

From https://docs.jax.dev/en/latest/_autosummary/jax.lax.custom_linear_solve.html

> `transpose_solve` ([Callable](https://docs.python.org/3/library/typing.html#typing.Callable)[[[Callable](https://docs.python.org/3/library/typing.html#typing.Callable), [Any](https://docs.python.org/3/library/typing.html#typing.Any)], [Any](https://docs.python.org/3/library/typing.html#typing.Any)] | None) – [...] **Required for backwards mode automatic differentiation, unless symmetric=True**, in which case solve provides the default value.

PR #405, opened 2025-04-11
Adding actuation model with TN curve 

CarlottaSartore:
----
📚 Documentation preview 📚: https://jaxsim--405.org.readthedocs.build//405/

### DIFF
--- a/src/jaxsim/api/model.py
+++ b/src/jaxsim/api/model.py
@@ -487,6 +487,7 @@ def reduce(
         terrain=model.terrain,
         contact_model=model.contact_model,
         contact_params=model.contact_params,
+        actuation_params=model.actuation_params,
         gravity=model.gravity,
         integrator=model.integrator,
     )
--- a/src/jaxsim/api/model.py
+++ b/src/jaxsim/api/model.py
@@ -57,6 +57,10 @@ class JaxSimModel(JaxsimDataclass):
         default=None, repr=False
     )
 
+    actuation_params: Static[jaxsim.rbda.actuation.ActuationParams] = dataclasses.field(
+        default=None, repr=False
+    )
+
     kin_dyn_parameters: js.kin_dyn_parameters.KinDynParameters | None = (
         dataclasses.field(default=None, repr=False)
     )
@@ -121,6 +125,7 @@ class JaxSimModel(JaxsimDataclass):
         terrain: jaxsim.terrain.Terrain | None = None,
         contact_model: jaxsim.rbda.contacts.ContactModel | None = None,
         contact_params: jaxsim.rbda.contacts.ContactsParams | None = None,
+        actuation_params: jaxsim.rbda.actuation.ActuationParams | None = None,
         integrator: IntegratorType | None = None,
         is_urdf: bool | None = None,
         considered_joints: Sequence[str] | None = None,
@@ -143,6 +148,7 @@ class JaxSimModel(JaxsimDataclass):
                 The contact model to consider.
                 If not specified, a soft contacts model is used.
             contact_params: The parameters of the contact model.
+            actuation_params: The parameters of the actuation model.
             integrator: The integrator to use for the simulation.
             is_urdf:
                 The optional flag to force the model description to be parsed as a URDF.
@@ -177,6 +183,7 @@ class JaxSimModel(JaxsimDataclass):
             time_step=time_step,
             terrain=terrain,
             contact_model=contact_model,
+            actuation_params=actuation_params,
             contact_params=contact_params,
             integrator=integrator,
             gravity=-gravity,
@@ -198,6 +205,7 @@ class JaxSimModel(JaxsimDataclass):
         terrain: jaxsim.terrain.Terrain | None = None,
         contact_model: jaxsim.rbda.contacts.ContactModel | None = None,
         contact_params: jaxsim.rbda.contacts.ContactsParams | None = None,
+        actuation_params: jaxsim.rbda.actuation.ActuationParams | None = None,
         integrator: IntegratorType | None = None,
         gravity: jtp.FloatLike = jaxsim.math.STANDARD_GRAVITY,
     ) -> JaxSimModel:
@@ -219,6 +227,7 @@ class JaxSimModel(JaxsimDataclass):
                 The contact model to consider.
                 If not specified, a relaxed-constraints rigid contacts model is used.
             contact_params: The parameters of the contact model.
+            actuation_params: The parameters of the actuation model.
             integrator: The integrator to use for the simulation.
             gravity: The gravity constant.
 
@@ -255,6 +264,9 @@ class JaxSimModel(JaxsimDataclass):
         if contact_params is None:
             contact_params = contact_model._parameters_class()
 
+        if actuation_params is None:
+            actuation_params = jaxsim.rbda.actuation.ActuationParams()
+
         # Consider the default integrator if not specified.
         integrator = (
             integrator
@@ -272,6 +284,7 @@ class JaxSimModel(JaxsimDataclass):
             terrain=terrain,
             contact_model=contact_model,
             contact_params=contact_params,
+            actuation_params=actuation_params,
             integrator=integrator,
             gravity=gravity,
             # The following is wrapped as hashless since it's a static argument, and we
--- a/src/jaxsim/api/model.py
+++ b/src/jaxsim/api/model.py
@@ -487,7 +487,6 @@ def reduce(
         terrain=model.terrain,
         contact_model=model.contact_model,
         contact_params=model.contact_params,
-        actuation_params=model.actuation_params,
         gravity=model.gravity,
         integrator=model.integrator,
     )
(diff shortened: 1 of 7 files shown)


younik (approved):
great!

traversaro (approved):

src/jaxsim/rbda/actuation/common.py
```diff
@@ -0,0 +1,11 @@
+import jaxsim.typing as jtp
+from jaxsim.utils import JaxsimDataclass
+
+class ActuationParams(JaxsimDataclass):
+    """
+    Abstract class representing the parameters of an actuation model.
```
flferretti, line 7: This is not an abstract class
```suggestion
    Parameters class for the actuation model.
```

src/jaxsim/api/model.py
```diff
@@ -272,6 +284,7 @@ def build(
             terrain=terrain,
             contact_model=contact_model,
             contact_params=contact_params,
+            actuation_params=actuation_params,
```
flferretti, line 287: This attribute should be passed also in `js.model.reduce` otherwise it won't be inherited

flferretti (commented):
Thanks @CarlottaSartore! I added some comments. By the way, I think that we could just use `api.actuation_model` module instead of creating a new one which may cause confusion

flferretti (changes requested):
By the way, I think that we could just use `api.actuation_model` module instead of creating a new one which may cause confusion

src/jaxsim/api/model.py
CarlottaSartore: Thanks, nice catch  ! Done in 49aa8db48affc36975abc0e2673898b41b05818e

CarlottaSartore:
About 

> By the way, I think that we could just use api.actuation_model module instead of creating a new one which may cause confusion

I have decided to create a new module because I thought of possible extensions in a near future with more complex actuation models, so I decided to replicate the same structure as for contact, which has an api.contact but also a module contacts where we implement the different contact models. I imagine the same for the actuation. 
What do you think @flferretti ?  Also @xela-95  with whom we created the first actuation class

CarlottaSartore:
> Thanks!
> 
> In future we could add simple unit tests in which we set a reference torque and check that this get limited by the actuation model.

Yes ! This could have sense ! I was struggling to identify a unit test to test this feature,  but what you are proposing could do the job !

CarlottaSartore:
> Thanks!
> 
> In future we could add simple unit tests in which we set a reference torque and check that this get limited by the actuation model.
> 
> Yes ! This could have sense ! I was struggling to identify a unit test to test this feature, but what you are proposing could do the job !

In the end I add the CI test in this PR  ! Feel free to review it , While doing it I noticed that: 

- We do not store in data the joint torque that we actually apply to the dynamics, it could have sense to store it   so that we can check the difference in between the desiredata and the actual torque applied. 
- I added a test for the actuation; we could also move the test for the joint limits since it is linked to the joint.  I did not do this in this PR since it would mean moving a few things around out of the scope of the PR,I added this comment only for for future reference

xela-95:
> In the end I add the CI test in this PR ! Feel free to review it , While doing it I noticed that:
> 
> * We do not store in data the joint torque that we actually apply to the dynamics, it could have sense to store it   so that we can check the difference in between the desiredata and the actual torque applied.
> * I added a test for the actuation; we could also move the test for the joint limits since it is linked to the joint.  I did not do this in this PR since it would mean moving a few things around out of the scope of the PR,I added this comment only for for future reference

Super!! Simple and effective as it should be! and yes, unit tests sometimes help to discover things we miss :)

flferretti:
> About
> 
> > By the way, I think that we could just use api.actuation_model module instead of creating a new one which may cause confusion
> 
> I have decided to create a new module because I thought of possible extensions in a near future with more complex actuation models, so I decided to replicate the same structure as for contact, which has an api.contact but also a module contacts where we implement the different contact models. I imagine the same for the actuation. What do you think @flferretti ? Also @xela-95 with whom we created the first actuation class

Alright! I was thinking that since the class is not in the `api`, it might be counter intuitive for the user to use that, but we can modify this behavior in the future since we have the same issue with the contact models. Thank you!

src/jaxsim/rbda/actuation/common.py
```diff
@@ -0,0 +1,12 @@
...
+
+
+class ActuationParams(JaxsimDataclass):
+    """
+    Parameters class for the actuation model.
+    """
+
+    τ_max: jtp.Float = 3000.0  # Max torque (Nm)
+    ω_th: jtp.Float = 30.0  # Threshold speed (rad/s)
+    ω_max: jtp.Float = 100.0  # Max speed for torque drop-off (rad/s)
```
flferretti, line 12: I'm usually a fan of unicode chars, but I'm super convinced of using them with user-exposed variables, e.g. in IPython these attribute would be difficult to access/create. Can we switch to ascii chars?

```suggestion
    max_torque: jtp.Float = 3000.0  # (Nm)
    threshold_speed: jtp.Float = 30.0  # (rad/s)
    dropoff_speed: jtp.Float = 100.0  # (rad/s)
```

flferretti, line 12: Can we use `dataclasses.field` as these attributes are type-annotated and so that they appear in `__repr__()`?
```suggestion
    τ_max: jtp.Float = dataclasses.field(default=3000.0)  # Max torque (Nm)
    ω_th: jtp.Float = dataclasses.field(default=30.0)  # Threshold speed (rad/s)
    ω_max: jtp.Float = dataclasses.field(
        default=100.0
    )  # Max speed for torque drop-off (rad/s)
```

flferretti (commented):
Last few comments, sorry

src/jaxsim/rbda/actuation/common.py
CarlottaSartore: Done in ba556cd4e97d43b10710ad06c9b3af842ddbcd8b

CarlottaSartore: done in ba556cd4e97d43b10710ad06c9b3af842ddbcd8b

CarlottaSartore: This actually  does not make the field instantiate correctly  see 
     
     ```
     """
        Compute the torque limits using the tn curve.
    
        Args:
            model: The model to consider.
            data: The data of the considered model.
    
        Returns:
            The torque limits.
        """
    
        τ_max = model.actuation_params.torque_max  # Max torque (Nm)
        ω_th = model.actuation_params.omega_th  # Threshold speed (rad/s)
        ω_max = model.actuation_params.omega_max  # Max speed for torque drop-off (rad/s)
        abs_vel = jnp.abs(data.joint_velocities)
        τ_lim = jnp.where(
>           abs_vel <= ω_th,
            τ_max,
            jnp.where(
                abs_vel <= ω_max, τ_max * (1 - (abs_vel - ω_th) / (ω_max - ω_th)), 0.0
            ),
        )
E       TypeError: '<=' not supported between instances of 'DynamicJaxprTracer' and 'Field'
E       --------------------
E       For simplicity, JAX has removed its internal frames from the traceback of the following exception. Set JAX_TRACEBACK_FILTERING=off to include these.

```

Do you know why this happens with fields ?

flferretti:
> @flferretti I should have implemented all the comments! Let me know if I can proceed with the merge !

Thanks a lot @CarlottaSartore! The error in CI can be fixed decorating the class with `@jax_dataclasses.pytree_dataclass`

src/jaxsim/rbda/actuation/common.py
flferretti: Sorry, I haven't seen this before writing https://github.com/ami-iit/jaxsim/pull/405#issuecomment-2797047047

CarlottaSartore: Should be done now !

flferretti:
@CarlottaSartore it looks like there is a slowdown of the `test_benchmark.py::test_relaxed_rigid_contact_model[128]`, can we wait a bit until we understand the reason? I don't think it's related to your PR though

CarlottaSartore:
> @CarlottaSartore it looks like there is a slowdown of the `test_benchmark.py::test_relaxed_rigid_contact_model[128]`, can we wait a bit until we understand the reason? I don't think it's related to your PR though

Yes sure ! I see that  the same happens here https://github.com/ami-iit/jaxsim/pull/407#pullrequestreview-2760079629. I will comment there also !

flferretti:
> > @CarlottaSartore it looks like there is a slowdown of the test_benchmark.py::test_relaxed_rigid_contact_model[128], can we wait a bit until we understand the reason? I don't think it's related to your PR though
> 
> 
> 
> There is no slowdown anymore, as can be seen here https://github.com/ami-iit/jaxsim/pull/405#pullrequestreview-2760098248, the slow down was most likely due to some CI problem rather than the PR itself. I think we can merge it now! What do you think @flferretti  ? 

I agree. Good to go for me, thanks!

CarlottaSartore:
Merging

PR #411, opened 2025-04-14
Implement custom JVP rule for improving `safe_norm` computation speed

flferretti:
This pull request introduces a custom JVP rule for the `safe_norm` function to handle NaNs and zero arrays more effectively during gradient computation, while having the same speed of the plain `jnp.linalg.norm` when the gradients are not necessary. This enhancement improves the runtime speed of the `safe_norm` function

Note that `@jax.custom_jvp` only supports functions with positional args or positional-or-keyword, for this reason the non-array args (axis, keepdims) should be trated as static. For this reason a wrapper factory was necessary

----
📚 Documentation preview 📚: https://jaxsim--411.org.readthedocs.build//411/

### DIFF
--- a/src/jaxsim/math/utils.py
+++ b/src/jaxsim/math/utils.py
@@ -1,6 +1,48 @@
+import functools
+
+import jax
 import jax.numpy as jnp
 
 import jaxsim.typing as jtp
+
+
+@functools.cache
+def _make_safe_norm(axis, keepdims):
+    @jax.custom_jvp
+    def _safe_norm(array: jtp.ArrayLike) -> jtp.Array:
+        """
+        Compute an array norm handling NaNs and making sure that
+        it is safe to get the gradient.
+
+        Args:
+            array: The array for which to compute the norm.
+
+        Returns:
+            The norm of the array with handling for zero arrays to avoid NaNs.
+        """
+        # Compute the norm of the array along the specified axis.
+        return jnp.linalg.norm(array, axis=axis, keepdims=keepdims)
+
+    @_safe_norm.defjvp
+    def _safe_norm_jvp(primals, tangents):
+        (x,), (x_dot,) = primals, tangents
+
+        # Check if the entire array is composed of zeros.
+        is_zero = jnp.all(x == 0.0)
+
+        # Replace zeros with an array of ones temporarily to avoid division by zero.
+        # This ensures the computation of norm does not produce NaNs or Infs.
+        array = jnp.where(is_zero, jnp.ones_like(x), x)
+
+        # Compute the norm of the array along the specified axis.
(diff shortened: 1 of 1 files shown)


younik (commented):
This is very smart, and the benchmarks look good!

I would maybe add a test for this.

> Note that @jax.custom_jvp only supports functions with positional args or positional-or-keyword, for this reason the non-array args (axis, keepdims) should be trated as static. For this reason a wrapper factory was necessary

I didn't undesrtand this issue. Why do we need the factory wrapper?

flferretti:
> This is very smart, and the benchmarks look good!

Thank you!

> I would maybe add a test for this.

There should be already a autodiff test in https://github.com/ami-iit/jaxsim/blob/b1cabaf93151b7ca9aefd818fcbd67c14b1939ea/tests/test_automatic_differentiation.py#L421-L462

Would you like to add anything to the test?
 
> > Note that @jax.custom_jvp only supports functions with positional args or positional-or-keyword, for this reason the non-array args (axis, keepdims) should be trated as static. For this reason a wrapper factory was necessary
> 
> I didn't undesrtand this issue. Why do we need the factory wrapper?

The reason is `@jax.custom_jvp` only supports positional arguments, and all non-array arguments must be static. In other words, JAX needs to be able to trace the function's behavior deterministically based on the array inputs only.

`axis` and `keepdims` are non-array arguments and they control how the function behaves (e.g., along which axis to compute the norm). If you just pass them normally to a `@custom_jvp` function, JAX would not know how to trace/compile the JVP properly

The wrapper factory like:

```python
@functools.cache
def _make_safe_norm(axis, keepdims):
    @jax.custom_jvp
    def _safe_norm(array):
        ...
    @_safe_norm.defjvp
    def _safe_norm_jvp(...):
        ...
    return _safe_norm
```

solves this by baking in the `axis` and `keepdims` values as **static**. So when you call:

```python
safe_norm(array, axis=1, keepdims=True)
```

under the hood it becomes:

```python
_make_safe_norm(axis=1, keepdims=True)(array)
```

Now JAX only sees a function with a single positional array argument, which is exactly what `@custom_jvp` expects. The `axis` and `keepdims` are static and outside the JAX trace

flferretti:
@ami-iit/darwin friendly ping :)

CarlottaSartore (approved):

xela-95 (approved):
Really neat code!! Thanks @flferretti

flferretti:
@ami-iit/darwin I removed the closure caching as I believe it can cause conflicts with JAX and potentially cause cache invalidation of JIT-compiled functions. I tried comparing with and without it with the benchmarks and they are better without the `@functools.cache`

PR #412, opened 2025-04-15
Compute the quaternion derivative by using the Hamilton product

flferretti:
This pull request includes several changes to the quaternion derivative calculations. The primary focus is on computing it by using the Hamilton product of quaternions, which allowed removing the `omega_in_body_fixed` parameter

Refactoring and simplification:

* [`src/jaxsim/math/quaternion.py`](diffhunk://#diff-03197f574252e3d07472d558e9b1e402e4a5139ee0d9f4d92a78818299561aafL71): Removed the `omega_in_body_fixed` parameter and refactored the `Quaternion.derivative` method to simplify the quaternion multiplication process. This includes the removal of the conditional logic for frame representation and the introduction of a more efficient computation using Einstein summation. [[1]](diffhunk://#diff-03197f574252e3d07472d558e9b1e402e4a5139ee0d9f4d92a78818299561aafL71) [[2]](diffhunk://#diff-03197f574252e3d07472d558e9b1e402e4a5139ee0d9f4d92a78818299561aafL80-R110)

Associated method calls updated:

* [`src/jaxsim/api/integrators.py`](diffhunk://#diff-33b49b4d51e5e0194d3ce2e4b4f53ef1901858b8ab341d88fa6c9620200094e1L55): Removed the `omega_in_body_fixed` parameter from the call to `Quaternion.derivative` in the `semi_implicit_euler_integration` function.
* [`src/jaxsim/api/ode.py`](diffhunk://#diff-5a992d6e02acddeb8cf23d21fa124750c80efd0234f9b3db840041fd7a602b29L131): Removed the `omega_in_body_fixed` parameter from the call to `Quaternion.derivative` in the `system_position_dynamics` function.
* [`tests/test_api_frame.py`](diffhunk://#diff-4e1c8d24898ae6710935ab409713452b512cef4991ba756700883b20166cd406L270): Removed the `omega_in_body_fixed` parameter from the call to `Quaternion.derivative` in the `compute_q̇` function.
* [`tests/test_api_link.py`](diffhunk://#diff-84fa59de8bb7185910d5e08365ac56136a256f30901dc0d30da30d8b48c3be49L360): Removed the `omega_in_body_fixed` parameter from the call to `Quaternion.derivative` in the `compute_q̇` function.
* [`tests/test_api_model.py`](diffhunk://#diff-4cda1019189e1ba59de5a3d73f5e78a64ca5e21bda7231ac2591c20bf377ae4bL461): Removed the `omega_in_body_fixed` parameter from the call to `Quaternion.derivative` in the `compute_q̇` function.

### DIFF
--- a/src/jaxsim/math/quaternion.py
+++ b/src/jaxsim/math/quaternion.py
@@ -68,7 +68,6 @@ class Quaternion:
     def derivative(
         quaternion: jtp.Vector,
         omega: jtp.Vector,
-        omega_in_body_fixed: bool = False,
         K: float = 0.1,
     ) -> jtp.Vector:
         """
--- a/src/jaxsim/math/quaternion.py
+++ b/src/jaxsim/math/quaternion.py
@@ -87,7 +87,17 @@ class Quaternion:
         # Construct pure quaternion: (scalar damping term, angular velocity components)
         ω_quat = jnp.hstack([K * safe_norm(ω) * (1 - safe_norm(quaternion)), ω])
 
-        # Apply quaternion multiplication based on frame representation
+        # Quaternion multiplication using index tables.
+        # This approach avoids using the explicit quaternion multiplication formula
+        # by encoding the necessary element-wise products and signs via indexed operations.
+        # Given two quaternions q and w, their Hamilton product q ⊗ w can be written
+        # as a combination of q[i] * w[j] terms with appropriate signs.
+        # i_idx and j_idx define which elements of the outer product q ⊗ w to select.
+        # For example, i_idx[1][2] = 2 and j_idx[1][2] = 3 means: take q[2] * w[3] for this term.
+        # sign_matrix[i][j] gives the sign (+1 or -1) to apply to each q[i] * w[j] term,
+        # depending on quaternion multiplication rules.
+        # This indexed summation reproduces the Hamilton product of quaternions in a
+        # vectorized way, and is suitable for use with JAX.
         i_idx = jnp.array([[0, 1, 2, 3], [0, 1, 2, 3], [0, 2, 3, 1], [0, 3, 1, 2]])
         j_idx = jnp.array([[0, 1, 2, 3], [1, 0, 3, 2], [2, 0, 1, 3], [3, 0, 2, 1]])
         sign_matrix = jnp.array(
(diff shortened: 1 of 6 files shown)


src/jaxsim/math/quaternion.py
```diff
@@ -77,59 +76,38 @@ def derivative(
...
-            true_fun=Q_body,
-            false_fun=Q_inertial,
-            operand=quaternion,
+        q = quaternion.squeeze()
+
+        # Construct pure quaternion: (scalar damping term, angular velocity components)
+        ω_quat = jnp.hstack([K * safe_norm(ω) * (1 - safe_norm(quaternion)), ω])
+
+        # Apply quaternion multiplication based on frame representation
+        i_idx = jnp.array([[0, 1, 2, 3], [0, 1, 2, 3], [0, 2, 3, 1], [0, 3, 1, 2]])
```
CarlottaSartore, line 91: This could be difficult to debug in the future, adding an explanation or a link to an explanation could be helpfull

CarlottaSartore (changes requested):
This PR changes the API which  should be used only internally by jaxsim, indeed from 

```
 jaxsim.math.Quaternion.derivative(
            quaternion=data.base_orientation,
            omega=W_ω_WB,
            omega_in_body_fixed=False,
        )
```

We pass to 
```
 jaxsim.math.Quaternion.derivative(
            quaternion=data.base_orientation,
            omega=W_ω_WB,
        )
        
  ```
  
  Let me allert @ami-iit/darwin

src/jaxsim/math/quaternion.py
flferretti: I'll add a more explanatory comment, thank you!

flferretti: Addressed in https://github.com/ami-iit/jaxsim/pull/412/commits/4f37c413e0cd18d9c128409ba086c76f9ce32829

CarlottaSartore (approved):

flferretti:
Friendly ping @ami-iit/darwin

xela-95 (approved):
Thanks @flferretti ! Just a question: I didn't understood if with this PR we made the choice of computing the quaternion derivative using only angular velocities only in body-fixed representation, since you dropped the conditional to handle the omega in inertial case (that is also the case for mixed representation of 6D velocities). 
Did this change has the goal of better support for differentiability of performance increase?

PR #443, opened 2025-06-05
Allow to override joint friction coefficients during parsing

flferretti:
This pull request introduces new configuration options for joint dynamics and updates the parser to utilize these options. It adds environment variables for static and viscous friction coefficients and modifying the parser to use these variables as defaults.

- No friction

https://github.com/user-attachments/assets/c6599401-7252-4fc3-b22c-09600771243c

- Viscous friction

https://github.com/user-attachments/assets/1ac73dae-8bde-422f-bacb-b71a469f2f06

- Static friction

https://github.com/user-attachments/assets/a3e528eb-01f5-42b5-89e6-75d3a428a14d

----
📚 Documentation preview 📚: https://jaxsim--443.org.readthedocs.build//443/

### DIFF
--- a/docs/guide/configuration.rst
+++ b/docs/guide/configuration.rst
@@ -47,10 +47,21 @@ Joint Dynamics
 ~~~~~~~~~~~~~~
 Joint dynamics are configured using environment variables starting with ``JAXSIM_JOINT_``. Available variables include:
 
+- ``JAXSIM_JOINT_FRICTION_STATIC``: Overrides the static friction coefficient for all joints.
+
+  *Default:* ``0.0``.
+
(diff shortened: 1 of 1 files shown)


docs/guide/configuration.rst
```diff
@@ -47,10 +47,21 @@ Joint Dynamics
 ~~~~~~~~~~~~~~
 Joint dynamics are configured using environment variables starting with ``JAXSIM_JOINT_``. Available variables include:
 
+- ``JAXSIM_JOINT_FRICTION_STATIC``: Overrides the static friction coefficient for all joints.
```
CarlottaSartore, line 50: Why the decision to put them as env variables?
 
These values should be dependent on the model. I am not convinced of having the default set via env variables.

PR #468, opened 2025-09-22
Update default contact model to Soft and remove depracated warning

CarlottaSartore:
In this PR, I have updated the default contact model to the soft one (since it is the one currently used the most). Updated the method documentation accordingly, and removed the deprecated warning for the hardware parameters


flferretti (approved):
LGTM, thanks!

xela-95:
rebased on main

